PR #215, opened 2019-09-05
Re-license under the GPLv3.

bartfeenstra:


bartfeenstra:
The old `LICENSE` file still exists in `setup.py`'s `data_files`.

Issue #253, opened 2019-10-29
The search results show incorrectly translated instructions


bartfeenstra:
The search results sometimes show incorrectly translated instructions.

PR #254, opened 2019-11-07
Localize the search instructions

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/253. One drawback is that if the container templates contain a backtick it will fail.


bartfeenstra:
Closing in favor of https://github.com/bartfeenstra/betty/pull/347

PR #335, opened 2020-02-20
WIP: Add sources and facts to citations and media pages

bartfeenstra:
## To-dos
- Turn label and meta macros into includable templates, so they can inherit contexs (a person context would be in a variable called `person`). The thing is that place meta needs a place context on top of the place subject, to determine how detailed the meta should be.
- Another option is to have every label/meta macro/template take a `context` variable that's a dict, keyed by context type.


bartfeenstra:
Closing in favor of https://github.com/bartfeenstra/betty/pull/337

Issue #351, opened 2020-03-06
Do not anonymize events if only non-subject presences are private


bartfeenstra:
Events at which private people are present in roles other than primary subject shouldn't be anonymized and cleaned. Only when one of the subjects is private should this happen.

PR #352, opened 2020-03-06
Add placeholders for private citations and sources

bartfeenstra:
Citations and sources are now privatized, anonymized, and cleaned from ancestries. This ensures privacy, but also removes the indication of the existence of supporting evidence for the facts presented on generated sites.

Instead of removing private citations and sources, replace them with anonymous citations and sources specifically designed to act as placeholders for the real citations and sources. These communicate supporting evidence exists, without giving away any private details.

This fixes https://github.com/bartfeenstra/betty/issues/351


betty/plugins/anonymizer/__init__.py
```diff
@@ -8,7 +8,34 @@
...
+        Citation.__init__(self, source)
+
+    @property
+    def location(self) -> Optional[str]:
+        return _("This citation has not been published in order to protect people's privacy")
+
+    def replace(self, other: Citation) -> None:
+        self.facts.append(*other.facts)
+        self.files.append(*other.files)
+        self.source = other.source
```
bartfeenstra: This line must be removed, as the containing source is *always* an `AnonymousSource`.

betty/plugins/anonymizer/__init__.py
```diff
@@ -8,7 +8,34 @@
 from betty.plugins.privatizer import Privatizer
 
 
+class AnonymousSource(Source):
+    def __init__(self):
+        Source.__init__(self, _('Private'))
+
+    def replace(self, other: Source) -> None:
+        self.citations.append(*other.citations)
+        self.contained_by = other.contained_by
```
bartfeenstra: `AnonymousSource` is never contained by another source. It's entirely separate.

PR #410, opened 2020-04-17
removed unused character in list place page template

terror:
cool project


bartfeenstra:
Hi @terror, thanks for the PR. It seems you accidentally added a bunch of IntelliJ files that need cleaning up.

PR #411, opened 2020-04-17
removed unused character from listplace page template

terror:
Ended up messing up that other branch somehow, here's the fix on a new branch without the intelliJ files. Sorry about that!


bartfeenstra:
Fabulous, thanks!

Issue #155, opened 2019-07-15
Split generation out per content type


bartfeenstra:
Move generation to plugins, and let plugins expose generators. A genrator can output public data and ahould then define a content define the formats (content types, etc) they can generate. By default we ship with a plugin that generates `text/html`, but we can later add others, such as `application/ld+json` , or even one for nginx configuration.

bartfeenstra:
Alternatively we can take a file-based approach. This means that for every resource type we want to render, there's a directory in the file system with templates. For each resource, these are copied to the resource's destination directory, and then rendered. The advantages are that the generation API does not need to know which files are being rendered: it just provides the source directory path, destination directory path, and template variables. The renderers will do the rest. Plugins and individual site configurations can then override or add files.

PR #458, opened 2020-05-05
Generate resources based on template directories.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/155


bartfeenstra:
Site generation is significantly slower than before, possibly due to more file I/O since we first copy all files to their destination, then render the tree which, for some renderers, skips their ability to render on copy.

bartfeenstra:
The problem is that this completely prevents us from adding fallback templates later, using `jinja2.environment.Environment.select_template()`.

Issue #483, opened 2020-08-28
Refactor the select_* filters to tests


bartfeenstra:
Refactor the se`lect_*` Jinja2 filters to tests instead.

bartfeenstra:
Closing because any such tests would need access to the context, which Jinja2 does not provide.

PR #484, opened 2020-08-28
Refactor the select_dateds and select_localizeds Jinja2 filters into tests

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/483

The issue is that tests cannot be context-aware.


bartfeenstra:
Closing because any such tests would need access to the context, which Jinja2 does not provide.

PR #554, opened 2020-11-05
Windows compatibility

bartfeenstra:


package.json
```diff
@@ -15,5 +15,10 @@
     "stylelint": "^13.6.1",
     "stylelint-config-standard": "^20.0.0",
     "stylelint-no-unsupported-browser-features": "^4.0.0"
+  },
+  "scripts": {
+    "cypress": "cypress",
```
bartfeenstra, line 20: Does this need to be declared explicitly, or can we call these executables through `npm run` without?

bartfeenstra: Yes, we need this apparently.

Issue #618, opened 2020-11-29
Lazily compile translations 


bartfeenstra:
Translations are now made and compiled, and then both the translation sources as well as their compiled versions are committed to Git. This causes some duplication, and because of the required compilation step, anyone wanting to add or override translations for their site will need to perform this step themselves. This is a major UX hurdle for our users who are not familiar with the command line. If we can ship with translation sources only and compile those on the fly, we allow users full control, yet reduce the number of tasks they will have to perform manually.

PR #608, opened 2020-11-28
Add French translations

patlx:
fr_FR-LC_MESSAGES renamed fr_FR/LC_MESSAGES

PR #619, opened 2020-11-29
Compile translations lazily

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/618

## To-dos
- Update the test so it covers the `*.po` files.
- If this is merged **before** https://github.com/bartfeenstra/betty/pull/608, update that PR to remove the `*.mo` file.
- If this is merged **after** https://github.com/bartfeenstra/betty/pull/608, remove the `*.mo` file in this PR.


betty/tests/assets/locale/test_translations.py
```diff
@@ -0,0 +1,15 @@
...
+
+_LOCALES_DIRECTORY_PATH = path.join(path.dirname(path.dirname(path.dirname(path.dirname(__file__)))), 'assets', 'locale')
+
+
+class TranslationsTest(TestCase):
+    def test(self) -> None:
+        locale_paths = glob.glob(path.join(_LOCALES_DIRECTORY_PATH, '**'))
+        for locale_path in locale_paths:
+            locale_directory_name = path.basename(locale_path)
+            with open(path.join(_LOCALES_DIRECTORY_PATH, locale_directory_name, 'LC_MESSAGES', 'betty.mo'), 'rb') as f:
```
bartfeenstra, line 14: This needs to test (compile) the *.po files now.

betty/locale.py
```diff
@@ -275,11 +280,27 @@ def negotiate_localizeds(preferred_locale: str, localizeds: List[Localized]) ->
 
 
 def open_translations(locale: str, directory_path: str) -> Optional[gettext.GNUTranslations]:
```
bartfeenstra, line 282: Break BC and make this private. Then in the future we can consider compiling all layered translations into one.

bartfeenstra: Let's not break BC here. We can always do that later if/when we've actually figured out how to aggregate translations and compile them into a single `*.mo` file.

Issue #642, opened 2020-12-07
Make the Configuration and Site reactive/observable


bartfeenstra:
We currently presume `Configuration` does not change once loaded into a `Site`, but some `Site` attributes do change during its lifetime, most notable when entering or exiting its context, and changing its locale. We currently hardcode the few supported use cases, but making these two objects entirely reactive would provide a unified approach that can replace the current hardcoded solutions as well as allow other changes to be made more freely. It would also help https://github.com/bartfeenstra/betty/pull/641 so the GUI can be updated based on state changes more easily.

PR #644, opened 2020-12-08
Add an observableproperty() decorator.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/642

Untested pseudo code at this time. Ideally we'd add similar decorators for collections such as dictionaries and lists, Though theoretically this one decorator would allow us to subclass these collection types and decorate each relevant method, that would be ugly and cumbersome, because these types have multiple methods that represent the same or highly similar actions, e.g. `list.__setitem__` and `list.append()`. If we are to add custom decorators for these collections, we must first decide on the semantics: do `list.__setitem__` and `list.append()` both represent the same 'set' action? If so, should we pass on the index of the inserted value when calling the observer?

### DIFF
--- a/betty/config.py
+++ b/betty/config.py
@@ -1,5 +1,5 @@
 import json
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 from os import path
 from typing import Dict, Type, Optional, List, Callable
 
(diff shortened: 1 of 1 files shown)


bartfeenstra:
Fixed by https://pypi.org/project/reactives/, which is based on code originally written for https://github.com/bartfeenstra/betty/pull/641, and then extracted into a Python package.

PR #728, opened 2021-05-16
Replace the graph API with a dependency on the new tsorted package.

bartfeenstra:
## Change log
- The `betty.graph` module was replaced with the [tsorted](https://pypi.org/project/tsorted) package.


bartfeenstra:
Explorer whether we shouldn't be using https://pypi.org/project/graphlib-backport/ instead.

bartfeenstra:
Closing this in favor of https://github.com/bartfeenstra/betty/pull/730

PR #778, opened 2021-09-06
Upgrade to Qt 6

bartfeenstra:


bartfeenstra:
This would break BC as the minimum operating system requirements have been increased significantly. For example, Qt5 runs on Ubuntu 18+, but Qt6 required Ubuntu 20+. I want to support OSs from more than the past year, so let's postpone this upgrade for now.

PR #797, opened 2021-12-20
Make the URL generators faster

bartfeenstra:
Profiling showed that Betty spends a relatively large portion of the time generating URLs. Speeding this up would have a relatively large positive impact on site generation time.

## Proposals 
- Add a method to URL generators that is a type guard and checks if a resource can be handled by that generator.
- For BC, provide q default implementation to that method. Add deprecation warnings when a generator raises a ValueError.
- Benchmark if using the job context's memory cache makes things faster.


bartfeenstra:
Can we instead use the future Cache API for this?

PR #894, opened 2022-07-04
Keep a memory-backed cache of Jinja2-filtered files and images  to prevent excessive I/O

### LINKED ISSUE #893: Cache image and file Jinja2 filters to reduce I/O
The `image` and `file` Jinja2 filters incur a lot of I/O due to some needless linking and copying, but also because even without that, the I/O involved with checking if a file was filtered already also slows site generation down.

Instead, can we cache filtered files, and clear the cache when site generation is done?

The implementation requires that at least one API knows about another. For example:
- the Jinja2 filters know about the site generation API, and set up their own cache in response to it
- the site generation API isolates the filters' cache bin(s) using a context manager, restoring the original cache when generation is done

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/893

### DIFF
--- a/betty/cache.py
+++ b/betty/cache.py
@@ -1,7 +1,8 @@
 import logging
 import shutil
+import threading
 from contextlib import suppress
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Iterator, Optional
 
 from betty import fs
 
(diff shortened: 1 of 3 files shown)


bartfeenstra:
This has been obsoleted by https://github.com/bartfeenstra/betty/pull/1049

PR #997, opened 2023-06-11
Make entities dumpable

bartfeenstra:
This lets us
- remove the `betty.json` module
- embed JSON-LD into HTML pages for improved SEO
- allow entities of third-party types to be rendered as JSON as well as the built-in entity types


bartfeenstra:
This was obsoleted by https://github.com/bartfeenstra/betty/pull/1224

PR #1010, opened 2023-07-15
Do not omit or remove data from ancestries when processing them

### LINKED ISSUE #495: Privatization breaks the relationships graph
Relationships between people are defined between parents and children only. This lets us derive information about people's partners, by finding out the other parents of people's children.

However, when children are privatized and cleaned, this breaks the relationships graph, and some people's partners can no longer be found. This has become even clearer after https://github.com/bartfeenstra/betty/issues/305, where partners are missing from people's timelines entirely.

### LINKED ISSUE #496: Limit people's associated events to their expected lifetime
https://github.com/bartfeenstra/betty/issues/305 introduced associated events to people's timelines. However, in the case of missing start-of-life or end-of-life events, associated events can go back or forward way beyond someone's birth or death.

Just like the `Privatizer` plugin uses a lifetime threshold (set to a 100 years), so should the associated events collection. In fact, we should have a single global configuration option for this, that can be used by Betty and any plugins.

### LINKED ISSUE #706: The Cleaner extension keeps files with citations
The Cleaner extension keeps files with citations. This is a left-over from when a file's citations and sources were actually resources using the file, whereas now these citations are references to sources the file was retrieved from.

## Definition of Done
- Make the Cleaner extension clean files even if they have citations
- Make the Cleaner extension clean citations even if they have files
- Make the Cleaner extension clean sources even if they have files
- Update the documentation (https://github.com/bartfeenstra/betty/pull/637)

### LINKED ISSUE #718: Allow the Cleaner to keep private people
Allow the Cleaner extension to keep private people. This lets site show the site of families and give a hint about the extent of the available data, without giving away people's private details. Do consider if merely indicating that a public person has a certain number of children or partners is not really a privacy infringement after all.

### LINKED ISSUE #1022: Flattened entity collections are slow
A ~10,000- entity ancestry takes a few seconds to be unflattened on my machine, and that's just when Gramps is done loading. We flatten and unflatten ancestries multiple times when generating a single site.

## Suggestions
- Consider making `FlattenedEntityCollection.unflatten()` return an iterator instead of an entity collection, as we often throw away that collection because we want the entities in an `Ancestry` instead.
- Consider allowing `Ancestry` to pause its on-add behavior when adding a collection of entities that we know is a full graph.
- Consider using `@lru_cache` on the entity association registry as we know it called thousands of times with only a few different argument values.
- Add a method to compare entities. See if comparing them by type and ID isn't faster than entire object comparisons. Then we can also use sets for fast lookups internally.
- Consider extracting the parts of `_FlattenedAssociation` into a new type that describes the entity type association (much like `EntityTypeAssociation`, but then pickleable). `_FlattenedAssociation` is then replaced by a tuple of owner and associate IDs, and a reference to this new type. This saves a string and a reference for each flattened association.
- One step further is to refactor `Entity` and association decorators, so they override `Entity.__getstate__` and replace the attribute values with a a single entity ID (for to-one associations) or a tuple of entity IDs (for to-many associations). Lowest memory footprint! This also addresses to bug where flattened entity collections will only ever guarantee a stable order for one side of any association (whichever one is encountered first), causing problems for bidirectional associations.
- Benchmark!

bartfeenstra:
Do not omit or remove data from ancestries when processing them, such as during derivation. To that end, remove the Anonymizer and Cleaner extensions.

This fixes https://github.com/bartfeenstra/betty/issues/495
This fixes https://github.com/bartfeenstra/betty/issues/496
This obsoletes https://github.com/bartfeenstra/betty/issues/706
This fixes https://github.com/bartfeenstra/betty/issues/718
This fixes https://github.com/bartfeenstra/betty/issues/1022


betty/deriver.py
```diff
@@ -190,19 +213,30 @@ def _get_derivable_events(person: Person, derivable_event_type: type[EventType])
...
         reference_event = presence.event
 
         if reference_event is None:
             continue
 
-        # We cannot reliably determine dates based on reference events with calculated date ranges, as those events
-        # would start or end *sometime* during the date range, but to derive dates we need reference events' exact
-        # start and end dates.
-        if isinstance(reference_event, DerivedEvent):
+        if reference_event.date is None:
```
bartfeenstra, line 228: We do nees to take fuzzy dates into account. For example, we must not derive a death that comes before a lifetime event.

PR #1031, opened 2023-09-08
Prevent the Jinja2 `file` and `image` filters from performing redundant tasks

### LINKED ISSUE #893: Cache image and file Jinja2 filters to reduce I/O

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/893
This obsoletes https://github.com/bartfeenstra/betty/pull/894

## To-dos
- ?

# Follow-ups
- Should generation tasks also receive a batch-scoped logger? We'd need a `QueueHandler` to make that multiprocessing-safe.
- make dispatcher operate on a task batch as well, if the caller wants to


betty/app/__init__.py
```diff
@@ -117,14 +110,22 @@ def dump(self) -> VoidableDump:
         }, True)
 
 
+@final
 class App(Configurable[AppConfiguration], ReactiveInstance):
+    _thread_pool: _TaskPool
```
bartfeenstra, line 115: This is no longer needed

bartfeenstra:
This has been obsoleted by https://github.com/bartfeenstra/betty/pull/1049

PR #1226, opened 2024-01-27
Remove multiprocessing

### LINKED ISSUE #1225: Remove multiprocessing
Does multiprocesing actually benefit Betty much now that everything is asynchronous and threaded? If we remove it, we no longer lose time pickling and can remove `Pickleable`.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/1225


betty/extension/nginx/serve.py
```diff
@@ -5,23 +5,23 @@
...
 
 
 class DockerizedNginxServer(AppServer):
     def __init__(self, app: App) -> None:
-        super().__init__(
-            # Create a new app so we can modify it later.
-            dill.loads(dill.dumps(app))
-        )
+        # Create a new project so we can modify it later.
+        project = Project()
```
bartfeenstra: This requires the existing ancestry as well

bartfeenstra: Also update the comment so it's clear we are doing this to update the project **configuration**

betty/extension/nginx/serve.py
```diff
@@ -5,23 +5,23 @@
...
 from betty.serve import NoPublicUrlBecauseServerNotStartedError, AppServer
 
 
 class DockerizedNginxServer(AppServer):
     def __init__(self, app: App) -> None:
-        super().__init__(
-            # Create a new app so we can modify it later.
-            dill.loads(dill.dumps(app))
-        )
+        # Create a new project configuration, so we can modify it later.
```
bartfeenstra: Can't we make the changes right here?

betty/generate.py
```diff
@@ -34,7 +31,7 @@
 from betty.string import camel_case_to_kebab_case, camel_case_to_snake_case
 from betty.task import Context
```
bartfeenstra, line 32: Task contexts should be internal to the generation API for now.

betty/app/__init__.py
```diff
@@ -348,7 +348,7 @@ def concurrency(self) -> int:
         with suppress(KeyError):
             return int(stdos.environ['BETTY_CONCURRENCY'])
         # Assume that any machine that runs Betty has at least two CPU cores.
-        return stdos.cpu_count() or 2
+        return (stdos.cpu_count() or 2) * 4
```
bartfeenstra: Decide what to do with this

PR #1238, opened 2024-02-03
Make Gramps loader messages translatable 

### LINKED ISSUE #1237: Make Gramps loader messages translatable
`GrampsLoader.load_tree()` in https://github.com/bartfeenstra/betty/blob/0.3.x/betty/gramps/loader.py#L138 logs several hardcoded messages. These should be translatable.

## Definition of Done
- Make all log messages translatable through the gettext methods o. `self._localizer`
- Remove the `@todo` comment

unixmiddleware:
Localized hard coded messages

### DIFF
--- a/betty/gramps/loader.py
+++ b/betty/gramps/loader.py
@@ -148,30 +148,29 @@ class GrampsLoader:
         database = self._tree.getroot()
 
         self._load_notes(database)
-        # @todo Localize all log messages
-        logger.info(f'Loaded {self._added_entity_counts[Note]} notes.')
+        logger.info(self._localizer._(f'Loaded {self._added_entity_counts[Note]} notes.'))
 
         self._load_objects(database, self._gramps_tree_directory_path)
-        logger.info(f'Loaded {self._added_entity_counts[File]} files.')
+        logger.info(self._localizer._(f'Loaded {self._added_entity_counts[File]} files.'))
 
         self._load_repositories(database)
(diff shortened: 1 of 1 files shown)


betty/gramps/loader.py
```diff
@@ -148,30 +148,29 @@ async def load_tree(self, tree: ElementTree.ElementTree, gramps_tree_directory_p
         database = self._tree.getroot()
 
         self._load_notes(database)
-        # @todo Localize all log messages
-        logger.info(f'Loaded {self._added_entity_counts[Note]} notes.')
+        logger.info(self._localizer._(f'Loaded {self._added_entity_counts[Note]} notes.'))
```
bartfeenstra, line 151: The way string translation works, is that all strings MUST be passed in to the gettext methods as static string literals. Literals because of the way they are extracted, static because the strings themselves act as unique keys.

In short, that means any formatting must be done after translation. The solution is straightforward: use regular strings instead of f-strings, and call `.format()` on the translated string to inject the variables. Use named instead of indexed placeholders to allow translators to change their order of appearance.

bartfeenstra (changes requested):
Thanks! Just got one comment and I believe the tests will inform us the translation files have not been updated by running `betty update-translations`.

bartfeenstra:
Hi @unixmiddleware ! Could you use a hand with this?

PR #1310, opened 2024-02-24
Add a cache API

bartfeenstra:
## To do
- ensure the versions mentioned in the deprecations are accurate
- Create follow-ups to remove deprecated functionality
  - Refactor `DocumentationServer` to receive a `BinaryFileCache`
  - Refactor `LocalizerRepository` to receive a `BinaryFileCache`


betty/cache.py
```diff
@@ -5,23 +5,210 @@
...
+from pickle import dumps, loads
+from typing import Sequence, Self, Generic, TypeVar, TypeAlias, Mapping
+
+import aiofiles.os
+from aiofiles.ospath import getmtime
 
-from betty import fs
 from betty.locale import Localizer
 
+CacheItemValue: TypeAlias = bytes | bool | int | float | str | None | Sequence['CacheItemValue'] | Mapping[str, 'CacheItemValue']
```
bartfeenstra: remove the collections so we can guarantee exact types better

betty/cache.py
```diff
@@ -5,23 +5,210 @@
...
+    async def get(self, id: str) -> CacheItem[CacheItemValue] | None:
+        cache_item = await self._get(id)
+        if cache_item is None and self._wraps is not None:
+            return await self._wraps.get(id)
+        return cache_item
+
+    async def _get(self, id: str) -> CacheItem[CacheItemValue] | None:
+        raise NotImplementedError
+
+    async def set(self, id: str, value: CacheItemValue) -> None:
```
bartfeenstra: Only call self._set() if there is no wrapped cache. Also clear self's cache upon set

betty/asyncio.py
```diff
@@ -78,3 +80,20 @@ async def run(self) -> None:
...
+
+    async def __aenter__(self):
+        while not self._threading_lock.acquire(False):
+            # Sleeping for zero seconds does not actually sleep, but gives the event
+            # loop a chance to progress other tasks.
+            await sleep(0)
+        await self._asyncio_lock.acquire()
+
+    async def __aexit__(self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None) -> None:
+        self._asyncio_lock.release()
```
bartfeenstra: This requires an await

betty/app/__init__.py
```diff
@@ -417,7 +417,7 @@ def servers(self) -> Mapping[str, Server]:
         }
 
     @property
-    def cache(self) -> Cache:
+    def cache(self) -> FileCache:
         if self._cache is None:
-            self._cache = Cache(self.localizer)
+            self._cache = FileCache(self.localizer, ASSETS_DIRECTORY_PATH)
```
bartfeenstra: Delete this when the app locale changes

betty/asyncio.py
```diff
@@ -78,3 +80,20 @@ async def run(self) -> None:
             self._return_value = await self._awaitable
         except BaseException as e:
             self._e = e
+
+
+class ThreadSafeLock:
```
bartfeenstra: Actually, a thread lock also works across async, so we can remove the asyncio.Lock from this class.

betty/cache.py
```diff
@@ -1,35 +1,319 @@
...
+    async def delete(self, id: str) -> None:
+        raise NotImplementedError
 
-class _Cache:
     async def clear(self) -> None:
         raise NotImplementedError
 
 
-class FileCache(_Cache):
+class NoOpCache(Cache[CacheItem]):
```
bartfeenstra: Remove this again as it's not used

betty/app/__init__.py
```diff
@@ -418,7 +419,11 @@ def servers(self) -> Mapping[str, Server]:
         }
 
     @property
-    def cache(self) -> FileCache:
-        if self._cache is None:
-            self._cache = FileCache(self.localizer, fs.CACHE_DIRECTORY_PATH)
-        return self._cache
+    def cache(self) -> Cache[CacheItem]:
```
bartfeenstra: This is a BC break! Revert this!

betty/cache.py
```diff
@@ -1,35 +1,287 @@
...
         localizer: Localizer,
-        cache_directory_path: Path,
-        /,
+        *,
+        scopes: Sequence[str] | None = None,
     ):
         self._localizer = localizer
+        self._scopes = scopes or ()
+        self._scoped_caches: dict[str, Self] = {}
+        self._lock = AsynchronizedLock(Lock())
```
bartfeenstra: Move this to subclasses. A no-op cache would not need a lock, for instance.

betty/tests/conftest.py
```diff
@@ -47,6 +51,15 @@ def set_logging(caplog: LogCaptureFixture) -> Iterator[None]:
         yield
 
 
+@pytest.fixture(scope='function')
+async def file_cache() -> AsyncIterator[FileCache]:
+    """
+    Patch Betty's default global file cache with a temporary directory.
```
bartfeenstra: Rewrite this and remove any mention of patching.

betty/tests/__init__.py
```diff
@@ -1,44 +1,33 @@
...
 from contextlib import asynccontextmanager
 from pathlib import Path
 from typing import Callable, TypeVar, Any, AsyncIterator, Awaitable, ParamSpec
 
 import aiofiles
 import html5lib
-from aiofiles.tempfile import TemporaryDirectory
 from html5lib.html5parser import ParseError
 from jinja2.environment import Template
+from typing_extensions import deprecated
```
bartfeenstra: Do we need to add this to pyproject.yml again?

betty/tests/__init__.py
```diff
@@ -1,44 +1,33 @@
...
 from betty.jinja2 import Environment
 from betty.json.schema import Schema
 from betty.locale import Localey
+from betty.tests.test_cli import _patch_cache
 
 T = TypeVar('T')
 P = ParamSpec('P')
 
 
+@deprecated('The `@patch_cache` decorator has been deprecated as of Betty 0.3.1. Use the `file_cache` fixture instead.')
```
bartfeenstra: 0.3.2!

betty/tests/test_cache.py
```diff
@@ -0,0 +1,144 @@
...
+
+import pytest
+from pytest_mock import MockerFixture
+
+from betty.cache import MemoryCache, FileCache
+from betty.locale import DEFAULT_LOCALIZER
+
+
+class TestFileCache:
+    async def test_scopes(self, tmp_path: Path) -> None:
```
bartfeenstra: Extract the test methods that cover `Cache` implementations into a test base class.

betty/cache.py
```diff
@@ -1,35 +1,343 @@
...
+    @property
+    def modified(self) -> int | float:
+        return self._modified
+
+
+class Cache(Generic[CacheItemT], _HasScopes):
+    def with_scope(self, scope: str) -> Self:
+        raise NotImplementedError
+
+    async def _lock(self, id: str) -> AsyncLock:
```
bartfeenstra: Use `cache_item_id` instead of `id` everywhere.

betty/wikipedia.py
```diff
@@ -110,53 +108,55 @@ class _Retriever:
...
         self._ttl = ttl
         self._http_client = http_client
         self._rate_limiter = RateLimiter(200)
         self._images: dict[str, Image | None] = {}
 
-    async def _request(self, url: str, extension: str | None = None) -> Any:
-        cache_file_path = self._cache_directory_path / hashlib.md5(url.encode("utf-8")).hexdigest()
-        if extension:
-            cache_file_path = cache_file_path.with_suffix(f'.{extension}')
+    async def _fetch(self, url: str) -> bytes | None:
```
bartfeenstra: Extract fetch functionality into a separate class, which may be private for now. Doing this should allow us to refine the exact cache requirements for it.

betty/tests/test_cache.py
```diff
@@ -0,0 +1,131 @@
...
+
+import pytest
+from aiofiles.tempfile import TemporaryDirectory
+from pytest_mock import MockerFixture
+
+from betty.cache import MemoryCache, FileCache, Cache, CacheItem
+from betty.locale import DEFAULT_LOCALIZER
+
+
+class CacheBaseTest:
```
bartfeenstra: Update the test methods to cover concurrency as well

betty/tests/test_cache.py
```diff
@@ -0,0 +1,131 @@
...
+        ('scopey', 'dopey'),
+    ])
+    async def test_clear(self, scopes: Sequence[str], tmp_path: Path) -> None:
+        async with self._new_sut(scopes=scopes) as sut:
+            await sut.set('id', 'value')
+            await sut.clear()
+            assert await sut.get('id') is None
+
+
+class TestFileCache(CacheBaseTest):
```
bartfeenstra: Test the additional `FileCache` methods

betty/cache.py
```diff
@@ -1,35 +1,456 @@
...
+class BinaryFileCacheItem(FileCacheItem[bytes]):
+    def _load_value(self, value_bytes: bytes) -> bytes:
+        return value_bytes
+
+    @property
+    def path(self) -> Path:
+        return self._cache._cache_item_file_path(self.id)
+
+
+@deprecated(f"This class is deprecated, and exists only for App.cache's backwards compatibility. Use {Cache} instead.")
```
bartfeenstra: Add the versions in which this was deprecated and in which it will be removed.

betty/cache.py
```diff
@@ -1,35 +1,456 @@
...
+class FileCache(Cache[CacheItem[Any], Any]):
```
bartfeenstra: This must not extend any classes, and instead declare its own `clear()` method. Anything else is a BC break.

betty/cache.py
```diff
@@ -1,35 +1,467 @@
...
+        """
+        raise NotImplementedError
+
+    async def get(self, cache_item_id: str) -> CacheItemCoT | None:
+        """
+        Get the cache item with the given ID.
+        """
+        raise NotImplementedError
+
+    async def set(self, cache_item_id: str, value: CacheItemValueContraT, *, modified: int | float | None = None) -> None:
```
bartfeenstra: If there were to return the newly set cache item, proxies can immediately cache that as well. This means that any subsequent get will hit the proxy cache rather than the underlying cache which will often incur I/O.

bartfeenstra: Or introduce a new `.getset()` method for this in case no proxies are used?

betty/cache.py
```diff
@@ -1,35 +1,467 @@
...
+
+    @property
+    def modified(self) -> int | float:
+        return self._modified
+
+
+CacheItemValueSetter: TypeAlias = Callable[[CacheItemValueT], Awaitable[None]]
+
+
+class Cache(Generic[CacheItemCoT, CacheItemValueContraT], _HasScopes):
```
bartfeenstra: Move implementations to new submodules so the main cache module just contains the interfaces.

betty/cache.py
```diff
@@ -1,35 +1,467 @@
...
+
+
+@deprecated(f"This class is deprecated as of Betty 0.3.2, and will be removed in Betty 0.4.x. It exists only for App.cache's backwards compatibility. Use {Cache} instead.")
+class FileCache:
+    @property
+    def path(self) -> Path:  # type: ignore[empty-body]
+        pass
 
     async def clear(self) -> None:
+        pass
```
bartfeenstra: Is this overridden for the App's BC compatibility cache?

betty/cache.py
```diff
@@ -1,35 +1,467 @@
...
+
+
+class _FileCache(_CommonCacheBase[_FileCacheItem[CacheItemValueT], CacheItemValueT], Generic[CacheItemValueT]):
+    """
+    Provide a cache that persists cache items on a file system.
+
+    This cache can manage any cache item value that is pickleable.
```
bartfeenstra: The pickle comment no longer belongs here. Move it to a better spot.

betty/concurrent.py
```diff
@@ -33,18 +49,40 @@ async def acquire(self) -> None:
     def release(self) -> None:
         self._lock.release()
 
+    @classmethod
+    def threading(cls) -> Self:
+        return cls(threading.Lock())
+
+
+class MultiLock(_AsyncLock):
```
bartfeenstra: Add test coverage

betty/concurrent.py
```diff
@@ -33,18 +49,40 @@ async def acquire(self) -> None:
...
+    """
+    Provide a lock that is backed by one or more other locks.
```
bartfeenstra: This is actually "zero or more", but removing this detail entirely may be best.

betty/cache/file.py
```diff
@@ -0,0 +1,177 @@
...
+    __slots__ = '_id', '_cache', '_modified'
+
+    def __init__(
+        self,
+        id: str,
+        cache: _FileCache[Any],
+        modified: int | float,
+    ):
+        self._id = id
+        self._cache = cache
```
bartfeenstra: Keeping the cache inside items makes it easy to introduce cyclic dependencies. Remove this?

betty/cache/file.py
```diff
@@ -0,0 +1,177 @@
...
+    __slots__ = '_value', '_value_loaded'
+
+    def __init__(
+        self,
+        id: str,
+        cache: _FileCache[Any],
+        modified: int | float,
+    ):
+        super().__init__(id, cache, modified)
+        self._value: CacheItemValueT
```
bartfeenstra: What if we don't cache this here?

betty/cache/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+CacheItemT = TypeVar('CacheItemT', bound=CacheItem[Any])
+CacheItemCoT = TypeVar('CacheItemCoT', covariant=True)
+
+
+CacheItemValueSetter: TypeAlias = Callable[[CacheItemValueT], Awaitable[None]]
+
+
+class Cache(Generic[CacheItemCoT, CacheItemValueContraT], _HasScopes):
```
bartfeenstra: Can we simplify this by being generic only over the value type?

bartfeenstra: This exists only for the binary file cache but it's proven not to be all that useful to get a cache item's path from the item (when there is no item).

betty/tests/cache/test___init__.py
```diff
@@ -0,0 +1,63 @@
+from collections.abc import Iterator
+from pathlib import Path
+from typing import Sequence, AsyncContextManager, Generic
+
+import pytest
+
+from betty.cache import Cache, CacheItemValueT, CacheItemT
+
+
+class CacheBaseTest(Generic[CacheItemT, CacheItemValueT]):
```
bartfeenstra: Test getset()

bartfeenstra: Test that concurrency is non-blocking.

betty/tests/cache/test_file.py
```diff
@@ -0,0 +1,46 @@
...
+    def _values(self) -> Iterator[Any]:
+        yield True
+        yield None
+        yield 123
+        yield 123.456
+        yield []
+        yield {}
+
+
+class TestBinaryFileCache(CacheBaseTest[CacheItem[bytes], bytes]):
```
bartfeenstra: Test getting a cache item file path

betty/app/__init__.py
```diff
@@ -111,12 +112,20 @@ def dump(self) -> VoidableDump:
         }, True)
 
 
+class _BackwardsCompatiblePickledFileCache(PickledFileCache[Any], FileCache):
```
bartfeenstra, line 117: Schedule this for removal

betty/job.py
```diff
@@ -5,16 +5,29 @@
...
+    @deprecated('This method is deprecated as of Betty 0.3.2, and will be removed in Betty 0.4.x. Use `Context.cache` instead.')
     def claim(self, job_id: str) -> bool:
         with self._claims_lock:
             if job_id in self._claimed_job_ids:
                 return False
             self._claimed_job_ids.add(job_id)
         return True
+
+    @property
+    def cache(self) -> Cache[Any]:
```
bartfeenstra, line 31: Document that this is a volatile cache for this job context only.

betty/app/__init__.py
bartfeenstra: And deprecate it.

betty/concurrent.py
```diff
@@ -2,49 +2,112 @@
...
+    """
+
+    __slots__ = '_locked', '_locks'
+
+    def __init__(self, *locks: _Lock):
+        self._locks = locks
+        self._locked = False
+
+    async def acquire(self, *, wait: bool = True) -> bool:
+        acquisitions = await gather(*(lock.acquire(wait=wait) for lock in self._locks))
```
bartfeenstra, line 81: If this returns tuples, we won't have to `zip()` afterwards.

bartfeenstra: But then we wouldn't easily be able to check if any acquisition failed. Never mind!

PR #1317, opened 2024-02-26
Add test coverage for enabling and disabling extensions in the Graphical User Interface

bartfeenstra:


betty/extension/cotton_candy/gui.py
```diff
@@ -26,7 +26,13 @@ def __init__(self, app: App, color: _ColorConfiguration, *args: Any, **kwargs: A
         self.setFixedWidth(24)
 
     def paintEvent(self, a0: QPaintEvent | None) -> None:
-        painter = QPainter(self)
+        # @todo On macOS, this sometimes errors with
+        # @todo "RuntimeError: wrapped C/C++ object of type _ColorConfigurationSwatch has been deleted".
+        # @todo Ignore the error until it is fixed.
+        try:
```
bartfeenstra: Is this still needed?

bartfeenstra: This might be caused by the same thing as the other segfaults we are seeing.

PR #1330, opened 2024-03-01
Support Notes for all entities that have them in Gramps

bartfeenstra:


betty/extension/cotton_candy/assets/templates/notes.html.j2
```diff
@@ -0,0 +1,14 @@
+{% set notes = notes | select('public') | list %}
```
bartfeenstra: Rename this section-notes.html.j2 and add aliases for similar templates as well. Or do that in a follow-up.

PR #1346, opened 2024-03-24
Improve the memory usage of entity collections

bartfeenstra:
## To do
- Create follow-ups to remove deprecated functionality
- Consider the DX impact of this, as merging this change requires hard references everywhere.

### DIFF
--- a/betty/model/__init__.py
+++ b/betty/model/__init__.py
@@ -491,7 +491,7 @@ class ToMany(Generic[OwnerT, AssociateT], ToManyEntityTypeAssociation[OwnerT, As
         setattr(
             owner,
             self._owner_private_attr_name,
-            SingleTypeEntityCollection[AssociateT](self.associate_type)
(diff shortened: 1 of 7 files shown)


bin/test-pytest
```diff
@@ -7,4 +7,4 @@ cd "$(dirname "$0")/.."
 echo 'Running pytest...'
 
 coverage erase
-PYTHONWARNINGS='error:::betty[.*]' pytest --cov --cov-append --cov-config=.coveragerc --no-cov-on-fail "$@"
+PYTHONWARNINGS='error:::betty' pytest --cov --cov-append --cov-config=.coveragerc --no-cov-on-fail "$@"
```
bartfeenstra: Remove `PYTHONWARNINGS` as we now handle conversion to errors using a test fixture.

betty/warnings.py
```diff
@@ -0,0 +1,16 @@
+import warnings
```
bartfeenstra: Add docstrings to everything in this file.

Issue #1145, opened 2024-01-09
Core component decoupling


bartfeenstra:
## Proposals
- Split extensions into app extensions and project extensions
- Make `Project` depend on `App` instead of vice versa, so we can have multiple projects open per app
- In the CLI API, merge `@app_command` and `@global_command`, and inject the app into any and every command. We may need a `@project_command` though.
- Add an interface for core components that allows them to be started or stopped. Because App, Project, and Extension all need resource management for their services.
- Add a method to (some?) component interface for registering event handlers. Seeing as handlers can only be set when a dispatcher is present, this means projects can only register with apps, and extensions should only register with projects. Therefore we need an unregister method as .
- Each component has a boot() and shutdown() method. These are called automatically when entering the component as a context.
- Refactor extensions into project extension plugins.
- Remove `App.new_from_app()`

## Blocked by
- The `0.4.x` branch being opened
- https://github.com/bartfeenstra/betty/issues/980
- https://github.com/bartfeenstra/betty/issues/1537

PR #1609, opened 2024-06-22
Remove `Configuration.on_change()`

bartfeenstra:
## To do
- Consider fixing https://github.com/bartfeenstra/betty/issues/1145 here as well
- Add a message to the App configuration GUI window that changes will not take effect until the app is reloaded?
- Add a method to `ProjectWindow` to trigger a project change.
- Ensure configuration widgets receive this event. How to propagate this simply and easily? Given grandparent, parent, and child widgets where only the grandparent and (grand)child are `ProjectAwareMixin`, how to set the new project?
- Add `Configuration.update_transaction()` (name tbd) that is a context manager

### DIFF
--- a/betty/app/__init__.py
+++ b/betty/app/__init__.py
@@ -3,9 +3,8 @@
 from __future__ import annotations
 
 import operator
-import weakref
 from concurrent.futures import Executor, ProcessPoolExecutor
-from contextlib import suppress, asynccontextmanager
+from contextlib import suppress, asynccontextmanager, AsyncExitStack
 from functools import reduce
 from graphlib import CycleError, TopologicalSorter
 from multiprocessing import get_context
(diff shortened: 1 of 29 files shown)


bartfeenstra:
```python
    def clone(self) -> Self:
        """
        Clone the configuration into a new instance.

        This is a deep clone/copy for any data structures that are copied by
        reference and are owned by this instance, including but not limited to:
        - other :py:class:`betty.config.Configuration` instances
        - standard Python data structures such as lists and dictionaries
        Data structures that are copied by reference by are **not** owned by this
        instance **MUST NOT** be cloned/copied.

        Clones **MUST** be identical. This means that any dependencies, including
        environment variables, **MUST** be injected.
        """
        raise NotImplementedError(repr(self))
```

PR #1628, opened 2024-06-29
Decouple core components

### LINKED ISSUE #1145: Core component decoupling

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/1145


betty/cli.py
```diff
@@ -29,28 +30,31 @@
...
+    from betty.project.__init__ import Project
+    from collections.abc import Coroutine, Sequence, Mapping
     from PyQt6.QtWidgets import QMainWindow
 
 
 _P = ParamSpec("_P")
 
 
-class CommandProvider:
+def discover_commands() -> Mapping[str, type[Command]]:
```
bartfeenstra, line 46: Test this!

betty/core.py
```diff
@@ -0,0 +1,49 @@
+from contextlib import AsyncExitStack
+from types import TracebackType
+from typing import Self, Any
+
+
+class CoreComponent:
```
bartfeenstra, line 6: Mark as `@internal`

betty/extension/cotton_candy/__init__.py
```diff
@@ -34,7 +34,7 @@
 from betty.model.ancestry import Event, Person, Presence, is_public, Subject
 from betty.model.event_type import StartOfLifeEventType, EndOfLifeEventType
 from betty.os import link_or_copy
-from betty.project import EntityReferenceSequence, EntityReference
+from betty.project.__init__ import EntityReferenceSequence, EntityReference
```
bartfeenstra, line 37: Fix these imports everywhere

Issue #1625, opened 2024-06-27
Fix configuration change propagation


bartfeenstra:
## To do
- Ensure that there is a way for any configuration that is changed through the GUI to take effect
  - app configuration
  - project configuration (including service reloading!)
- Fix all `@todo` comments referencing `https://github.com/bartfeenstra/betty/issues/1625`
- Clone Project for jobs, eg load and generate

PR #1629, opened 2024-06-29
Decouple `App` and `Project`

### LINKED ISSUE #1145: Core component decoupling

bartfeenstra:
## Follow-ups
- Make Extension implement CoreComponent
- https://github.com/bartfeenstra/betty/issues/1625


betty/gui/project.py
```diff
@@ -743,12 +757,14 @@ def _navigate_to_pane(self, pane_name: str) -> None:
...
-        self._app.project.configuration.autowrite = True
+        self._bootstrap()
+        self._project.configuration.autowrite = True
         super().show()
 
     @override
     def close(self) -> bool:
-        self._app.project.configuration.autowrite = False
+        self._project.configuration.autowrite = False
+        wait_to_thread(self._async_exit_stack.aclose())
```
bartfeenstra, line 767: Set self._bootstrapped=False

betty/core.py
```diff
@@ -0,0 +1,62 @@
...
+
+from contextlib import AsyncExitStack
+from types import TracebackType
+from typing import Self, Any
+
+from betty.typing import internal
+
+
+@internal
+class CoreComponent:
```
bartfeenstra, line 13: Document that this class is internal but its public methods are, well, public. Maybe use a decorator for that too? Can the decorators add messages to the docstring?

PR #1645, opened 2024-07-05
Simplify `Server` and remove its dependency on aiohttp

bartfeenstra:


betty/serve.py
```diff
@@ -125,24 +127,25 @@ async def assert_available(self) -> None:
...
         Assert that this server is available.
 
         If this method returns, the server is considered available.
         If this method raises an exception, the server is considered unavailable.
         """
-        async with session.get(self.public_url) as response:
-            assert response.status == 200
+        url = urlparse(self.public_url)
+        connection = HTTPConnection(url.netloc)
+        await to_thread(connection.request, "HEAD", url.path)
```
bartfeenstra: Use GET so we test that the page content itself is also available.

betty/serve.py
```diff
@@ -113,24 +118,25 @@ async def assert_available(self) -> None:
...
-
-    async def _assert_available(self, session: ClientSession) -> None:
+        try:
+            await Do[Any, None](self._assert_available).until()
+        except Exception as error:
+            raise UserFacingError(
+                _("The server was unreachable after starting.")
+            ) from error
+
+    async def _assert_available(self) -> None:
```
bartfeenstra, line 135: Make this synchronous and run the whole method in a single thread.

PR #1709, opened 2024-07-14
Make site generation use multiprocessing

### LINKED ISSUE #1516: Remove entity association finalization
Entity association finalization was used when ancestries were still taken apart and reassembled as part of pickling them to other processes.

First consider if we really do not want to reintroduce multiprocessing 

## Definition of Done
- Deprecate
  - `_EntityTypeAssociation.finalize()`
  - `EntityTypeAssociationRegistry.finalize()`

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/1516

## To do
- Add `betty.typing.processsafe()` which implies `betty.typing.threadsafe()`
- Allow reduced (pickleable) values to implement a factory interface so they can he rehydrated after unpickling
- Do all pickling in one go so we don't risk pickling the same things twice
- Remove caches from job contexts (because memory caches cannot be multiprocess-safe), but introduce a new lightweight alternative, some kind of registry, just to do bookkeeping? This is also partly why performance right now is sub-optimal, as the image filters use this duplicates memory cache and therefore end up duplicating work between processes.
- **ACTUALLY** caches were designed so cache item values MUST be pickleable.... This should be documented.
- We should allow extensions to be 'pickled' as well, in order to share services between processes. For the Wikipedia extension, this would be its Wikipedia API rate limiter. Document this on the Extension class.
- Allow KeyboardInterrupt to gracefully shutdown the executor
- Should Jinja2 filters use the job pool?
- do we still need App's process pool then?
- Can we extract the pool functionality into a separate API/submodule? Even if we don't reuse this now, splitting the pool code from site generation reduces complexity and makes for easier testing. **Update**: we've now run into this in practice, where we want to mock something (testing third-party entity types), but obviously those mocks don't carry over to the processes in the pool. Separating the pool from the generation API means that we can test the pool fully (not needing any Project or Ancestry related mocking), site generation fully (by running the relevant tasks in the main process), and integration (by fully testing integration points **and nothing more**, because this is the part where conflicts occur)
- if we embed the pool in the job context, the context can do bookkeeping (prevent duplicates) for every task if needed (opt-in?)


betty/cache/_base.py
```diff
@@ -53,9 +53,9 @@ def __init__(
         self._scopes = scopes or ()
         self._scoped_caches: dict[str, Self] = {}
         self._locks: MutableMapping[str, _Lock] = defaultdict(
-            AsynchronizedLock.threading
+            AsynchronizedLock.multiprocessing
```
bartfeenstra: Confirm we end up using this.

betty/generate/pool.py
```diff
@@ -0,0 +1,205 @@
...
+    async def __aenter__(self) -> Self:
+        concurrency = os.cpu_count() or 2
+        # Avoid `fork` so as not to start worker processes with unneeded resources.
+        # Settle for `spawn` so all environments use the same start method.
+        executor = ProcessPoolExecutor(
+            max_workers=concurrency, mp_context=multiprocessing.get_context("spawn")
+        )
+        self._exit_stack.enter_context(executor)
+        # @todo Ensure we pass on the necessary dependencies
+        pickled_app_args = dill.dumps(
```
bartfeenstra: Can we not make the App directly pickleable? It is only Project that has a tricky dependency in the form of App

bartfeenstra: No, not if this would require Dill!

betty/generate/__init__.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+async def _log_jobs_forever(app: App, process_pool: _GenerationProcessPool) -> None:
+    with suppress(CancelledError):
+        while True:
+            await _log_jobs(app, process_pool)
+            await sleep(5)
+
+
+async def create_file(path: Path) -> AsyncContextManager[AsyncTextIOWrapper]:
```
bartfeenstra: Extract these helpers to a submodule?

betty/generate/__init__.py
```diff
@@ -0,0 +1,162 @@
...
+from betty.project import ProjectEvent
+
+if TYPE_CHECKING:
+    from betty.app import App
+    from betty.generate.pool import _GenerationProcessPool
+
+
+class GenerateSiteEvent(ProjectEvent):
+    """
+    Dispatched to generate (part of) a project's site.
```
bartfeenstra: Remove "part of"

betty/concurrent.py
```diff
@@ -81,6 +82,13 @@ async def acquire(self, *, wait: bool = True) -> bool:
     def release(self) -> None:
         self._lock.release()
 
+    @classmethod
+    def multiprocessing(cls) -> Self:
```
bartfeenstra, line 83: Test this

betty/generate/__init__.py
```diff
@@ -0,0 +1,162 @@
...
+            completed_job_percentage=floor(
+                completed_job_count / (total_job_count / 100)
+                if total_job_count > 0
+                else 0
+            ),
+        )
+    )
+
+
+async def _log_jobs_forever(app: App, process_pool: _GenerationProcessPool) -> None:
```
bartfeenstra: Make this part of the pool

betty/generate/pool.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+_GenerationProcessPoolTaskP = ParamSpec("_GenerationProcessPoolTaskP")
+
+worker_setup: Callable[[], None] | None = None
+
+
+class _GenerationProcessPool:
+    """
+    Set up a worker process, before the worker starts performing tasks.
```
bartfeenstra: Incorrect doc?

betty/__init__.py
```diff
@@ -2,6 +2,7 @@
 
 from pathlib import Path
 
+from betty import _patch  # noqa: F401
```
bartfeenstra: Separate this out into its own PR

betty/_patch.py
```diff
@@ -1,19 +1,28 @@
...
-    loop: BaseEventLoop, *args: Any, **kwargs: Any
-) -> None:
-    try:
-        await _original_shutdown_default_executor(loop, *args, **kwargs)
-    except RuntimeError as error:
-        # Work around a bug in Python 3.12 that will randomly cause a RuntimeError with the
-        # following message to be raised.
-        if "can't create new thread at interpreter shutdown" not in str(error):
-            raise
+if not _patched:
```
bartfeenstra: Separate this out into a separate PR

betty/cache/_base.py
```diff
@@ -47,14 +46,7 @@ def __init__(
...
         self._scoped_caches: MutableMapping[str, Self] = {}
-        self._locks: MutableMapping[str, _Lock] = defaultdict(
-            AsynchronizedLock.threading
-        )
-        self._locks_lock = AsynchronizedLock.threading()
-
-    async def _lock(self, cache_item_id: str) -> _Lock:
-        async with self._locks_lock:
-            return self._locks[cache_item_id]
+        self._lock = AsynchronizedLock.multiprocessing()
```
bartfeenstra: Separate this out into a separate PR

betty/concurrent.py
```diff
@@ -16,6 +17,13 @@
 
 from betty.asyncio import gather
 
+# @todo What is a good production value?
+# @todo The problem is perhaps (but maybe it is) not so much how many files we open in our own code
+# @todo but DEFINITELY the number of multiprocessing locks we use, each of which uses a pipe, which is a file handle
+# @todo
+# @todo Do we still need this?
+MAX_OPEN_FILES = multiprocessing.Manager().BoundedSemaphore(512)
```
bartfeenstra: We port this to processes in the generation pool, but not other processes. What about other threads?

betty/_patch.py
bartfeenstra: Call the file bootstrap.py?

betty/cache/_base.py
bartfeenstra: Also, benchmark what's better: a single lock, or lazily loading and deleting locks per cache item ID

betty/concurrent.py
bartfeenstra: Remove this. It's useless because of the differences between operating systems

betty/__init__.py
bartfeenstra: See https://github.com/bartfeenstra/betty/pull/1894

betty/model/collections.py
```diff
@@ -41,9 +41,6 @@ class EntityCollection(Generic[_TargetT], ABC):
 
     __slots__ = ()
 
-    def __init__(self):
```
bartfeenstra, line 44: Clean this and similar initializers in a separate PR

PR #1832, opened 2024-08-01
Migrate more generic functionality into static translations

bartfeenstra:


betty/locale/localizable/config.py
```diff
@@ -5,73 +5,30 @@
...
-        """
-        super().__init__()
-        self._minimum = minimum
-        self._translations: dict[str, str]
-        if translations is not None:
-            self.set(translations)
-        else:
-            self._translations = {}
-
-    def __getitem__(self, locale: str) -> str:
```
bartfeenstra, line 52: Any coverage we should remove?

PR #1836, opened 2024-08-02
Make link labels localizable

bartfeenstra:
This also coincidentally refactors the entire linked data and JSON Schema APIs


betty/json/linked_data.py
```diff
@@ -14,6 +16,19 @@
     from betty.model.ancestry import Link
 
 
+def carry_definitions(
```
bartfeenstra: Should we instead introduce a `Schema` class to manage this? Then we could fold the schema name into it for referenceable schemas.

betty/json/schema.py
```diff
@@ -4,77 +4,150 @@
...
+    def schema(self) -> DumpMapping[Dump]:
         """
-        Build the JSON Schema.
+        The raw JSON Schema.
         """
-        from betty import model
+        return self._schema
+
+    @property
+    def definitions(self) -> DumpMapping[Dump]:
```
bartfeenstra, line 43: Document that we only support top-level definitions

betty/json/schema.py
```diff
@@ -4,77 +4,150 @@
...
 
 class Schema:
     """
-    Build JSON Schemas for a Betty application.
+    A JSON Schema.
     """
 
-    def __init__(self, project: Project):
-        self._project = project
+    def __init__(
```
bartfeenstra, line 28: Add kwargs to set on the schema

PR #1932, opened 2024-09-05
Introduce a Factory API

### LINKED ISSUE #1920: Instance factories
- calling code will need to know if they can instantiate a plugin, and how
- How do we build it so calling code won't have to know about the exact factory method?
- IndependentFactory, ProjectDependentFactory, ...
- raise a helpful error as a fallback when instantiating using __init__ directly
- use for the Gramps loader

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/1920

### DIFF
--- a/betty/factory.py
+++ b/betty/factory.py
@@ -1,0 +1,33 @@
+"""
+Functionality for creating new class instances.
+"""
+
+from __future__ import annotations
+
+from typing import TypeVar, Self
+
+
+class FactoryError(RuntimeError):
+    """
+    Raised when a class could not be instantiated by a factory API.
+    """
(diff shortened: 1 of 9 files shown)


betty/factory.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def new(cls: type[_T]) -> _T:
+    """
+    :raises FactoryError: raised when the class could not be instantiated.
+    """
+    try:
+        return cls()
+    except Exception as error:
+        raise FactoryError from error
```
bartfeenstra: Add a friendly and helpful message.

PR #1964, opened 2024-09-16
Add genders

### LINKED ISSUE #901: More specific description based on genders
With the few small issues I solved, my website is starting to look good! However, one large issue remains, as I am aiming for a French version. The gender neutral description really does not work and people will not understand a sentence like "Ce sont les enfants de ..." ("They are the child of ...") or "Ils ont grandi avec un frère ou une sœur." ("They grew up with a sibling."). 

The neutral "they" does not exist in French, and there is no word for gender neutral sibling in French ("un frère ou une sœur" sounds like we do not know, which is clearly not true as the information is on the next column).

I had a look to change that, but it seems like you do not even load the gender information from Gramps, is that correct? It is not used anywhere? Would you be against loading it, so that we can know this information and adapt the string to be displayed when the gender is known?

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/901


documentation/development/plugin/gender.rst
```diff
@@ -0,0 +1,67 @@
...
+==============
+
+.. list-table::
+   :align: left
+   :stub-columns: 1
+
+   * -  Type
+     -  :py:class:`betty.ancestry.gender.Gender`
+   * -  Repository
+     -  :py:class:`betty.ancestry.gender.GENDER_REPOSITORY`
```
bartfeenstra: Not true! This must be Project.genders. Correct for other plugin types too, in separate PRs!

PR #1972, opened 2024-09-18
Add types for required and optional associations

### LINKED ISSUE #1962: Make associations being required/optional a part of their types
The reason we never did this before was when building a flattened ancestry, entities and associations are built separately and then merged. Meaning we must be able to build entities without associations. What about sentinels? Or allow associations to be lazy-loaded?

## To do
- ensured associations can be resolved in one go. An API function?

## Blocked by
- https://github.com/bartfeenstra/betty/pull/1966

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/1962

## To do
- Look through each template and remove conditions that are no longer needed because some attributes are now explicitly marked as required associations
- Add required validation for to-many associations, maybe even to collections?
- Resolvers are great at solving our typing problem. However how will they work for bidirectional associations? We don't always know how to set the inverse (which is what associations between two real entities do automatically, but resolvers cannot until resolved). E.g. if A has a resolvable value to B, B won't get *anything* that implies it should point back to A, until A's resolved entity is resolved, the entity set, and the associations inverting the association so B to A it set as well.


betty/model/association.py
```diff
@@ -5,35 +5,80 @@
...
+    """
+    An object that can resolve to another value when called.
+    """
+
+    @abstractmethod
+    def __call__(self) -> _T:
+        """
+        Return the resolved value.
+
+        :raises FailedToResolve: Raised if somehow resolving the value failed.
```
bartfeenstra: Remove "somehow". That sounds discouraging...

betty/model/association.py
```diff
@@ -5,35 +5,80 @@
...
+        """
+        pass
+
+
+Resolvable: TypeAlias = _T | Resolver[_T]
```
bartfeenstra: Split into ResolvableEntity and ResolvableEntityCollection

betty/model/association.py
```diff
@@ -5,35 +5,80 @@
...
+    pass
+
+
+class Resolver(Generic[_T], ABC):
+    """
+    An object that can resolve to another value when called.
+    """
+
+    @abstractmethod
+    def __call__(self) -> _T:
```
bartfeenstra: Rename to resolve() because we don't aim for Callable compatibility anyway

betty/model/association.py
```diff
@@ -5,35 +5,129 @@
...
+    def resolve(self) -> _T:
+        """
+        Return the resolved entity or entities.
+
+        :raises ResolutionError: Raised if resolution failed.
+        """
+        pass
+
+
+class RequiredEntityResolver(Generic[_EntityT], _Resolver[_EntityT]):
```
bartfeenstra: Do we really need the required/optional distinction here? If we return Entity it will satisfy any Entity|None.

betty/model/association.py
```diff
@@ -311,72 +423,216 @@ def new_attr(self, instance: _OwnerT & Entity) -> EntityCollection[_AssociateT]:
...
+        _OwnerT,
+        _AssociateT,
+        Intersection[_AssociateT, Entity],
+        RequiredResolvableEntity[Intersection[_AssociateT, Entity]],
+    ],
+):
+    _required = True
+
+    @override
+    def new_attr(self, instance: _OwnerT & Entity) -> Never:
```
bartfeenstra, line 405: new_attr() is required by Attr but not needed for required to-one relationships. Should we remove the method, and by extension the entire Attr API?

PR #2014, opened 2024-09-20
Add types for required and optional associations

### LINKED ISSUE #1962: Make associations being required/optional a part of their types

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/1962


betty/ancestry/__init__.py
```diff
@@ -21,16 +25,17 @@ def __init__(self):
         super().__init__()
         self._check_graph = True
 
-    def add_unchecked_graph(self, *entities: Entity) -> None:
+    @contextmanager
+    def unchecked(self) -> Iterator[None]:
```
bartfeenstra, line 29: Extract this into a separate PR

betty/model/association.py
```diff
@@ -17,16 +17,60 @@
...
 from betty.typing import internal
 
+_T = TypeVar("_T")
 _EntityT = TypeVar("_EntityT", bound=Entity)
 _OwnerT = TypeVar("_OwnerT")
 _AssociateT = TypeVar("_AssociateT")
 
 
-class Association(Generic[_OwnerT, _AssociateT]):
+class AssociationRequired(RuntimeError):
```
bartfeenstra, line 25: Do we still need this? It should only be raised in one place anyway

betty/model/association.py
```diff
@@ -17,16 +17,60 @@
...
+
+class ResolutionError(RuntimeError):
+    """
+    Raised when a :py:class:`betty.model.association.Resolver` cannot successfully resolve itself.
+    """
+
+    pass
+
+
+class Resolver(Generic[_T], ABC):
```
bartfeenstra: Should this be private?

PR #2037, opened 2024-09-25
Fix a bug where `wait_to_thread()` would await coroutines using the wrong event loops

bartfeenstra:


betty/asyncio.py
```diff
@@ -31,20 +39,24 @@ async def gather(*coroutines: Coroutine[Any, None, _T]) -> tuple[_T, ...]:
     return tuple(task.result() for task in tasks)
 
 
-def wait_to_thread(f: Awaitable[_T]) -> _T:
-    """
-    Wait for an awaitable in another thread.
-    """
-    synced = _WaiterThread(f)
+def wait_to_thread(
```
bartfeenstra, line 42: Extract this API change into a separate PR

PR #2074, opened 2024-10-05
Remove `wait_to_thread()`

### LINKED ISSUE #2057: Remove `wait_to_thread()`

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2057

## Ideas
- Allow assertions (and chains) to be synchronous and/or asynchronous? Chains must be typed so that synchronous chains cannot be extended by asynchronous assertions. The other way around is type-safe, and we don't have to worry about that.
- The problem is that some sync APIs (__getitem__ et al) use assertions for input validation. These cannot be made async like regular setters.

### DIFF
--- a/betty/app/config.py
+++ b/betty/app/config.py
@@ -9,13 +9,7 @@ from typing import final, Self
 from typing_extensions import override
 
 from betty import fs
-from betty.assertion import (
-    assert_record,
-    OptionalField,
-    assert_str,
-    assert_setattr,
-    assert_locale,
-)
+from betty.assertion import assert_record, OptionalField, assert_str, assert_locale
 from betty.config import Configuration
 from betty.serde.dump import Dump, minimize, DumpMapping
 from betty.typing import void_none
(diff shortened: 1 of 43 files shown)


betty/app/config.py
```diff
@@ -44,19 +38,22 @@ def locale(self) -> str | None:
         """
         return self._locale
 
-    @locale.setter
-    def locale(self, locale: str) -> None:
-        self._locale = assert_locale()(locale)
+    async def set_locale(self, locale: str) -> str:
```
bartfeenstra: Return str | None as per the getter

betty/assertion/__init__.py
```diff
@@ -41,13 +42,19 @@
...
+    Awaitable[_AssertionReturnT],
 ]
 
 _AssertionsExtendReturnT = TypeVar("_AssertionsExtendReturnT")
 _AssertionsIntermediateValueReturnT = TypeVar("_AssertionsIntermediateValueReturnT")
 
 
+_ChainableAssertion: TypeAlias = (
+    Assertion[_AssertionValueT, _AssertionReturnT]
+    | Callable[[_AssertionValueT], _AssertionReturnT]
```
bartfeenstra: Missing Awaitable

PR #2146, opened 2024-10-19
Make `Configurable` require `Configuration`

bartfeenstra:


betty/config/__init__.py
```diff
@@ -46,20 +47,32 @@ class Configurable(Generic[_ConfigurationT]):
...
 
 
+class DefaultConfigurable(Configurable[_ConfigurationT], Generic[_ConfigurationT]):
+    """
+    A configurable type that can provide its own default configuration.
+    """
+
+    @classmethod
+    @abstractmethod
+    async def default_configuration(cls) -> _ConfigurationT:
```
bartfeenstra: Prefix with `new_` to communicate that each invocation must return a new instance.

betty/plugin/config.py
```diff
@@ -131,3 +138,85 @@ def load_item(self, dump: Dump) -> PluginConfiguration:
     @classmethod
     def _create_default_item(cls, configuration_key: str) -> PluginConfiguration:
         return PluginConfiguration(configuration_key, {})
+
+
+async def ensure_plugin_configuration(
```
bartfeenstra: Raise a user facing error if the extension is configurable, cannot produce its own default config, and no config was given.

bartfeenstra: Make sure this is tied into config validation/loading

PR #2180, opened 2024-11-01
Update the Ukrainian translation

Illia-M:
Update ukrainian translation


bartfeenstra:
Looks great, and right in time for the beta. Duzhe dyakuyu!

PR #2188, opened 2024-11-03
Fix missing template files for entity list pages

### LINKED ISSUE #2187: Cotton Candy's file list page fails on missing template file
See https://github.com/Illia-M/test-betty-tree/actions/runs/11649767464/job/32437676231#logs

As originally reported by @Illia-M

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2187

## Changes
- Provide a fallback entity list page for any entity type
- Add tests to assert that all Cotton Candy entity list page templates function correctly 
- Fix the bug where the file list page errored

### DIFF
--- a/betty/project/extension/cotton_candy/assets/templates/entity/label--place.html.j2
+++ b/betty/project/extension/cotton_candy/assets/templates/entity/label--place.html.j2
@@ -5,12 +5,16 @@
     {%- if names | length == 0 -%}
         {%- set names = place.names -%}
     {%- endif -%}
-    {%- set name = (names | first).name | localize -%}
+    {%- if names -%}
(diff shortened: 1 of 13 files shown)


bartfeenstra:
@Illia-M Can you confirm this solves the problem you ran into?

Illia-M:
@bartfeenstra as I see problem with file not found - but as I see file should be inside gramps package...
https://github.com/Illia-M/test-betty-tree/actions/runs/11670539377/job/32494935155#step:6:97

bartfeenstra:
Can you report each individual problem as a new issue? Then it is easier to keep track of, and fix them one by one 😊

Illia-M:
Of course, sorry

PR #2194, opened 2024-11-05
Support relative file/media paths when loading Gramps family trees

### LINKED ISSUE #2193: Support relative media paths for Gramps family trees
@Illia-M I believe this is the problem you ran into, where your family tree uses relative paths to reference media files on your disk.

https://gramps.discourse.group/t/default-media-path-storage/79/3 indicates the necessary information to handle this automatically is included in the Gramps exports. If that is (still) true, we should be able to fix this with a couple of lines added to `GrampsLoader`.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2193


bartfeenstra:
@Illia-M This PR is mostly done, at least to the point where I believe it fixes the problem for you.

The remaining test failure is in case the Gramps export does not actually include a media directory path, but there are files/media with relative paths. I am hoping to learn more about this at https://gramps.discourse.group/t/relative-paths-without-mediapath/6383.

Illia-M:
@bartfeenstra as I see, now it try build full path to file on my PC and as expected - it fails
https://github.com/Illia-M/test-betty-tree/actions/runs/11699095682/job/32580473723

bartfeenstra:
Could it be that the file exists in your family tree at a relative path, but it does not live within the "base path for relative media paths" configured in your Gramps family tree? In that case, that would mean your family tree is incorrectly configured in Gramps itself: file paths must either be absolute, or be relative **to this base path**.

Illia-M:
Знімок екрана 2024-11-07 о 12 01 54

Illia-M:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/im/Documents/Gramps/Ukrainian-Example-Media/Історія_України-Русі._1912._Князь_Володимир_Великий.jpg'

/Users/im/Documents/Gramps/Ukrainian-Example-Media/ + Історія_України-Русі._1912._Князь_Володимир_Великий.jpg

bartfeenstra:
That screenshot does not show the media base path. That's a different setting in your family tree's "General" preferences pane.

Illia-M:
That only one place with path setting.
Screen of my General settings
image

PR #2309, opened 2025-02-11
Allow extensions to provide navigation links

### LINKED ISSUE #2286: Allow extensions to provide navigation links
Allow extensions to provide primary and secondary navigation links

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2286


test.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import annotations
```
bartfeenstra: Remove this entire test case file

bartfeenstra: Add it to a mypy bug report first!

PR #2376, opened 2025-02-27
Pickleable ancestry

bartfeenstra:
We cannot just pickle an entire large ancestry, for the simply reason that at some point this will run into Python's recursion limit. We should focus on flattening the ancestry graph before pickling.

## To do
- Error when pickling associations.

### DIFF
--- a/betty/tests/ancestry/test_citation.py
+++ b/betty/tests/ancestry/test_citation.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from typing import Sequence, Mapping, Any, TYPE_CHECKING
+import pickle
+from typing import Sequence, Mapping, Any, TYPE_CHECKING, cast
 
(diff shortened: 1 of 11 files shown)


betty/tests/model/test_association.py
```diff
@@ -2,6 +2,7 @@
 
 from typing import TypeVar, TYPE_CHECKING
 
+import dill
```
bartfeenstra: Remove the changes to this file

pyproject.toml
```diff
@@ -19,6 +19,7 @@ dependencies = [
     'aiohttp ~= 3.10, != 3.11.0',
     'asyncclick ~= 8.1',
     'babel ~= 2.16',
+    'dill ~= 0.3.9',
```
bartfeenstra: Undo this

PR #2391, opened 2025-03-01
Make the caches multiprocessing-safe

bartfeenstra:


betty/test_utils/cache/_base.py
```diff
@@ -0,0 +1,21 @@
...
+
+@internal
+class CommonCacheBaseTestBase(
+    Generic[_CacheItemValueT], CacheTestBase[_CacheItemValueT]
+):
+    """
+    A base class for tests of :py:class:`betty.cache._base._CommonCacheBase` implementations.
+    """
+
+    pass
```
bartfeenstra: Introduce a new test base for processafe caches and extend that here

betty/cache/__init__.py
```diff
@@ -46,6 +46,8 @@ class Cache(Generic[_CacheItemValueContraT], ABC):
 
     Implementations MUST be thread-safe.
 
+    Implementations SHOULD be multiprocessing-safe.
```
bartfeenstra: Use decorators

PR #2414, opened 2025-03-03
Allow configuration to be marked immutable

### LINKED ISSUE #2411: Allow configuration to be locked

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2411

### DIFF
--- a/betty/service.py
+++ b/betty/service.py
@@ -21,6 +21,7 @@ from typing import overload, TypeVar
 from warnings import warn
 
 from betty.concurrent import AsynchronizedLock, Lock
+from betty.config import Configurable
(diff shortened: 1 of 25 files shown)


betty/service.py
```diff
@@ -184,6 +185,8 @@ async def bootstrap(self) -> None:
         """
         self.assert_not_bootstrapped()
         self._bootstrapped = True
+        if isinstance(self, Configurable):
```
bartfeenstra, line 191: Unlock configuration in the shutdown

bartfeenstra: Add test coverage

PR #2417, opened 2025-03-06
Allow entities to be marked immutable

### LINKED ISSUE #2412: Allow entities to be locked
## To do
- Lock and unlock entities during project bootstrap and shutdown respectively. Except that we cannot load the ancestry until after project bootstrap... Do this in `betty.project.load.load()` instead?
- make plugin instances set on entities lockable as well, and propagate the locks from their owning entities
- allow collections to lock all their entities (primary propagation mechanism)

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2412

### DIFF
--- a/betty/project/load.py
+++ b/betty/project/load.py
@@ -40,6 +40,7 @@ async def load(project: Project) -> None:
     await project.event_dispatcher.dispatch(LoadAncestryEvent(job_context))
     await project.event_dispatcher.dispatch(PostLoadAncestryEvent(job_context))
     await _fetch_link_titles(project)
+    project.ancestry.immutable()
 
 
 async def _fetch_link_titles(project: Project) -> None:
--- a/betty/test_utils/model/collections.py
+++ b/betty/test_utils/model/collections.py
@@ -35,6 +35,12 @@ class EntityCollectionTestBase(Generic[_EntityT]):
         """
         raise NotImplementedError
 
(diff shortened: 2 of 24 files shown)


betty/project/load.py
```diff
@@ -40,6 +40,7 @@ async def load(project: Project) -> None:
     await project.event_dispatcher.dispatch(LoadAncestryEvent(job_context))
     await project.event_dispatcher.dispatch(PostLoadAncestryEvent(job_context))
     await _fetch_link_titles(project)
+    project.ancestry.immutable()
```
bartfeenstra, line 44: Add test coverage

betty/test_utils/model/collections.py
```diff
@@ -0,0 +1,42 @@
...
+from collections.abc import Sequence
+from typing import TypeVar, Generic
+
+from betty.model import Entity
+from betty.model.collections import EntityCollection
+
+_EntityT = TypeVar("_EntityT", bound=Entity)
+
+
+class EntityCollectionTestBase(Generic[_EntityT]):
```
bartfeenstra: Extract this into a separate PR

bartfeenstra: Document this on the collection class

bartfeenstra: See https://github.com/bartfeenstra/betty/pull/2419

PR #2549, opened 2025-04-14
Improve featured entity layouts

### LINKED ISSUE #2422: Optimize featured entities' card widths
Optimize features entities' card widths so
- cards can be landscape as well as portrait
- the orientation is determined based on the card's width using container size queries
- what do do with images if card content is really tall in landscape orientation?

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2422

## To do
- Make cards container-aware, with e.g. a `.card-group` element? Then they know how big their container is and how wide to make themselves and which orientation to use.
- Maybe add variables for card breakpoints, e.g. when to be portrait and when landscape. Using those, containers know how to lay out the cards. (This is for image cards only)


betty/project/extension/raspberry_mint/assets/templates/-front-featured-entities.html.j2
```diff
@@ -1,7 +1,7 @@
 <div class="container">
```
bartfeenstra: If we use .container without .row, we can simplify our code a bit

PR #2565, opened 2025-04-23
Fix a bug where the Gramps loader would error erroneously when parsing newer Gramps XML versions

### LINKED ISSUE #2560: Betty does not work with Gramps XML 1.7.2 (Gramps 6.0.0)
The gramps loader hard-codes the namespace to `http://gramps-project.org/xml/1.7.1/` so things start to break once the file use something like gramps 1.7.2.

Probably it would be best to check the actual namespace used in the file instead of hard-coding it.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2560

This can be installed by running `pip install git+https://github.com/bartfeenstra/betty.git@gramps-xml-compatibility`.

### DIFF
--- a/betty/assets/locale/ar/betty.po
+++ b/betty/assets/locale/ar/betty.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Betty VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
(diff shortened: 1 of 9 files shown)


bartfeenstra:
@FelixSchwarz Would you be able to test if this works for your Gramps files?

FelixSchwarz:
I wasn't sure how to install the version from source without polluting my system but I applied the patch on top of betty 0.4.15 does indeed solve the problem.

Issue #2583, opened 2025-05-01
Wikipedia image copyright is Wikimedia contributors


bartfeenstra:
The Wiki plugin has grown from just fetching Wikipedia article summaries to fetching much more, also from Wikimedia Commons, from which is where it fetches images. We currently erroneously state that these images are **Wikipedia** contributors, but in fact Wikimedia Commons' query API exposes information about the individual who owns the copyright.

## To do
- Create a new copyright plugin for a single Wikimedia contributor and that has attributes to configure that person's name and a URL to the copyright statement.
  - fetch the author's name from the query API
  - In `Client`, when getting an image, set the new copyright plugin and fill it's author and URL attributes.

## Follow-up issue/PR
- backport to 0.4.x

jbh1996:
Hey Bart! I know we talked earlier on a different issue. Do you have code already written for this issue? If not I'd like to try to solve this issue.

bartfeenstra:
@jbh1996 Hiya! No code for his one yet, so it's free for the taking. I'll update the OP with some more details.

Follow the [installation instructions](https://betty.readthedocs.io/stable/installation/#from-source) to set up a development environment. You can test your changes locally by running `bin/test` if you want fast feedback.

PR #2644, opened 2025-05-20
Fix incorrect copyright notice for Wikimedia images

jbh1996:
This pull request creates the copyright class in the wiki client.py and updates the get_image function to populate it. It also edits the Image class to include a copyright attribute.


bartfeenstra:
@jbh1996 Thanks for your PR! As you can see, some tests are failing. You can run these locally with `bin/test`. Most static check fails can usually be fixed automatically by running `bin/fix`. This is a good idea to do regularly during development as it informs you of errors that are often easy to fix, such as type errors or code style violations.

What's the scope of this PR? To display this copyright information on the file pages? E.g. to fix https://github.com/bartfeenstra/betty/issues/2583 ?

betty/wiki/client.py
```diff
@@ -57,7 +59,16 @@ class Image:
...
+    copyright: Copyright | None
 
+@final
+@dataclass(frozen=True)
+class Copyright:
+    """
+    Copyright information for a Wikimedia Commons contributor.
+    """
+    author: str
+    copyright_url: str
```
bartfeenstra: Let's just call this `url` for brevity.

betty/wiki/client.py
```diff
@@ -186,9 +197,24 @@ async def get_image(self, page_language: str, page_name: str) -> Image | None:
                         f"Could not successfully parse the JSON content returned by {url}: {error}"
                     )
                 ) from error
+
+            # Fetch copyright information
+            copyright = None
```
bartfeenstra: If you do this in the `except` clause, it's clearer why this is set to `None`. You'll probably have to declare the variable's full type beforehand either way.

betty/wiki/client.py
```diff
@@ -186,9 +197,24 @@ async def get_image(self, page_language: str, page_name: str) -> Image | None:
...
+            try:
+                artist = image_info["extmetadata"]["Artist"]["value"]
+                # Strip any HTML tags from the artist field, as it may contain markup
+                clean_artist = re.sub(r"<[^>]+>", "", unescape(artist)).strip()
```
bartfeenstra: Can we use html5lib for that? It's already a Betty dependency, and HTML is complex enough we don't want to do this ourselves and get it wrong.

betty/wiki/client.py
```diff
@@ -186,9 +197,24 @@ async def get_image(self, page_language: str, page_name: str) -> Image | None:
...
+                if clean_artist:
+                    copyright = Copyright(
+                        author=clean_artist,
+                        copyright_url=f"{image_info['descriptionurl']}#Licensing",                    )
+            except (LookupError, KeyError):
```
bartfeenstra: `KeyError` extends `LookupError` so we don't need to catch both.

bartfeenstra (changes requested):

bartfeenstra:
@jbh1996 I noticed your Betty fork was made from Betty 0.4.x. Since then, the 0.5.x branch was created and all new development happens there (with bug fixes potentially backported to 0.4.x once fixed in 0.5.x). You'll want to make sure your local development branch is kept up to date with 0.5.x, or it'll end up diverging so much your Git changes can no longer be applied (preventing a PR merge). Let me know if you need help with something :+1:

jbh1996:
Hey Bart! I just got a chance to work on this again. I made some edits and the bin/fix is giving me no errors back. Let me know if there's still changes that need to be made to the pull request. Sorry this took me a bit to get to.

bartfeenstra:
@jbh1996 No worries, and take your time. Thanks for addressing the feedback and fixing the test failures. Your code looks good but we're now at the point where we need to do something with this new functionality, as well as write tests for it. In order to determine exactly what else needs to be done, is it https://github.com/bartfeenstra/betty/issues/2583 you are trying to fix? If so (please confirm in a comment), then what remains is:
- To create a new [copyright plugin](https://betty.readthedocs.io/stable/development/plugin/copyright-notice/) that lets Betty assign this newly fetched data to actual file entities. Place the new plugin in `betty.wiki.copyright_notice`, along with the existing one.
- Update `betty.wiki.populator.Populator._image_file_reference()`, which is the function responsible for converting the data returned by the Wikipedia `Client` to entities for in the data model. You'll see it currently already sets a copyright notice, but that's the wrong one.
- Because we no longer need `betty.wiki.populator.Populator._copyright_notice`, the entire attribute may be removed, the corresponding argument removed from `__init__()`, and calling code updated so it no longer provides this argument.

jbh1996:
Yes it is https://github.com/bartfeenstra/betty/issues/2583 that I'm trying to fix. I'll get into those files in the next few days to try to make this work. Thanks for your patience and pointers.

jbh1996:
I've made changes to both the copyright_notice and populator file for your review!

betty/wiki/client.py
```diff
@@ -57,6 +58,18 @@ class Image:
     title: str
     wikimedia_commons_url: str
     name: str
+    image_copyright: Copyright | None
```
bartfeenstra, line 61: The class is called "image" already, so we can simply call the attribute "copyright".

betty/wiki/populator.py
```diff
@@ -244,7 +241,15 @@ async def _image_file_reference(
...
                     links=links,
-                    copyright_notice=self._copyright_notice,
+                    copyright_notice=(
+                        await WikimediaCommonsCopyright.new_for_app(
+                            self._ancestry.app,
+                            author=image.image_copyright.author,
+                            url=image.image_copyright.url,
+                        )
+                        if image.image_copyright
+                        else await WikipediaContributors.new_for_app(self._ancestry.app)
```
bartfeenstra, line 251: The original issue this PR tries to fix is that we are currently telling people these images are copyright Wikipedia contributors, but that is not the case (the images do not even come from Wikipedia). If we cannot get copyright information from Wikimedia, we should therefore let the image copyright be empty.

betty/wiki/populator.py
```diff
@@ -244,7 +241,15 @@ async def _image_file_reference(
                     path=image.path,
                     media_type=image.media_type,
                     links=links,
-                    copyright_notice=self._copyright_notice,
+                    copyright_notice=(
+                        await WikimediaCommonsCopyright.new_for_app(
```
bartfeenstra, line 245: These factory methods exist so that code can instantiate classes without knowing what arguments their `__init__()` methods require, usually because the class name is not hardcoded. In this case we know the class name, so we can instantiate it without having to call `.new_for_app()`.

bartfeenstra (commented):
Please have a look at the failing tests. For fastest feedback, run the tests locally on your own machine before pushing, so you can fix failures quickly without having to wait for the Github builds to complete.

PR #2663, opened 2025-06-11
Support automatically exporting family trees from Gramps

### LINKED ISSUE #2660: Allow the Gramps extension to export Gramps family trees automatically
Gramps comes with a command line that allows the export of family trees. The Gramps extension can then load these exports into an ancestry.

## Definition of Done
- Instead of a file name, allow the Gramps configuration to alternatively use family tree names
- When loading a family tree by name, spawn a subprocess to make Gramps export the file to a temporary directory, from which it is then loaded as usual
- See https://github.com/bartfeenstra/ancestry/blob/main/bin/export for the Gramps command to export a family tree

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2660

### DIFF
--- a/betty/gramps/loader.py
+++ b/betty/gramps/loader.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import gzip
 import re
+import sys
 import tarfile
 from collections import defaultdict
 from contextlib import ExitStack, suppress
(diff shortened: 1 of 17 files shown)


betty/gramps/loader.py
```diff
@@ -298,6 +304,7 @@ def __init__(
             str, Callable[[], PresenceRole | Awaitable[PresenceRole]]
         ]
         | None = None,
+        gramps_executable: Path | None = None,
```
bartfeenstra: Add this to `GrampsConfiguration`

bartfeenstra: Also document the new configuration option

PR #2699, opened 2025-06-28
Make `LocalizerRepository`'s API and usage synchronous

### LINKED ISSUE #2705: Remove `LocalizerRepository.locales`
Remove `LocalizerRepository.locales`, because any locale supported by the system is supported here.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2705

## To do
- We've got a pickling problem. When preloading services with async factories, there is no way to do this again upon unpickling. Not unless we enforce a bootstrap of the unpickled service provider, which we probably do not want.
- ensure full test coverage of the new APIs


betty/service.py
```diff
@@ -412,36 +418,79 @@ def _get(self, instance: _ServiceProviderT) -> _ServiceT:
...
+    *,
+    shared: bool = False,
 ) -> _SynchronousServiceManager[_ServiceProviderT, _ServiceT]:
     pass
 
 
 @overload
-def service(factory: None = None, shared: bool = False) -> _ServiceDecorator:
+def service(
+    factory: None = None,
```
bartfeenstra, line 481: What if we remove support for no factories, and enforce `service` as a decorator only?

betty/service.py
```diff
@@ -412,36 +418,79 @@ def _get(self, instance: _ServiceProviderT) -> _ServiceT:
...
+    factory: Callable[[_ServiceProviderT], _ServiceT],
+    *,
+    shared: bool = False,
 ) -> _SynchronousServiceManager[_ServiceProviderT, _ServiceT]:
     pass
 
 
 @overload
-def service(factory: None = None, shared: bool = False) -> _ServiceDecorator:
+def service(
```
bartfeenstra, line 480: What if we disallow the decorator from being called with a factory directly? Would that simplify our overloads for this to pass type checking?

betty/service.py
```diff
@@ -397,9 +410,28 @@ async def _get(self, instance: _ServiceProviderT) -> _ServiceT:
             return new_service
 
 
+class _AsynchronousFactorySynchronousGetterServiceManager(
+    Generic[_ServiceProviderT, _ServiceT],
+    ServiceManager[_ServiceProviderT, _ServiceT, Awaitable[_ServiceT], _ServiceT],
+):
+    async def initialize(self, instance: _ServiceProviderT) -> None:
```
bartfeenstra: Rename to `preload()`

betty/service.py
```diff
@@ -397,9 +410,28 @@ async def _get(self, instance: _ServiceProviderT) -> _ServiceT:
...
+        # @todo DO NOT DO THIS if there is an override instance.
```
bartfeenstra: Only do this if `self._get_attr() is not Void`

betty/service.py
```diff
@@ -412,62 +444,84 @@ def _get(self, instance: _ServiceProviderT) -> _ServiceT:
...
+    Decorate a synchronous service factory method.
 
     The factory method is replaced with a :py:class:`service manager <betty.service.ServiceManager>` which handles lazy
     service instantiation, caching, and multiprocessing support.
 
-    The decorated factory method should return a new service instance.
+    The decorated factory method must return a new service instance.
     """
 
     def _service(
```
bartfeenstra, line 662: Return a lambda here too

betty/service.py
bartfeenstra: Add test coverage

betty/locale/translation/project/__init__.py
```diff
@@ -0,0 +1,53 @@
+"""
+Manage translations for projects.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from betty.locale.translation import (
+    _new_translation,
```
bartfeenstra, line 10: Don't import private symbols.

PR #2741, opened 2025-07-29
Fix project translations

### LINKED ISSUE #2704: Prevent `mo` files from being compiled during tests
## Definition of Done 
- Fix this for App.translations
- Fix this for Project.translations. Use a decorator around App.translations instead of using App.assets directly

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2704 and this fixes a bug where projects would provide built-in app-level translations even if their apps were configured with different translations.

### DIFF
--- a/betty/assets.py
+++ b/betty/assets.py
@@ -4,25 +4,116 @@ The Assets API.
 
 from __future__ import annotations
 
+from abc import ABC, abstractmethod
 from asyncio import to_thread
+from contextlib import suppress
 from os import walk
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Self
+
+from typing_extensions import override
 
 from betty.concurrent import AsynchronizedLock
 from betty.typing import threadsafe
 
 if TYPE_CHECKING:
-    from collections.abc import AsyncIterator, Mapping, Sequence
+    from collections.abc import AsyncIterable, Iterable, Mapping, Sequence
+
+
+class AssetError(Exception):
+    """
+    Raised for asset API errors.
+    """
+
+
+class UnknownAsset(AssetError):
+    """
+    Raised when a requested asset cannot be found.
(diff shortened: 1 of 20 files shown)


betty/assets.py
```diff
@@ -4,27 +4,98 @@
...
+                if path not in seen:
+                    seen.add(path)
+                    yield path
+
+    @override
+    async def get(self, path: Path) -> Path:
+        for upstream in self._upstreams:
+            with suppress(Exception):
+                return await upstream.get(path)
+        # @todo Add a specific error class for this case
```
bartfeenstra: Fix this

betty/locale/translation/__init__.py
```diff
@@ -160,11 +168,110 @@ def find_source_files(
...
+
+    @abstractmethod
+    def get(self, locale: Localey) -> gettext.NullTranslations:
+        """
+        Get the translations for the given locale.
+        """
+
+
+@final
+class NoOpTranslationRepository(TranslationRepository):
```
bartfeenstra, line 191: Remove this in favor of StaticTranslationRepository

betty/locale/translation/__init__.py
```diff
@@ -23,13 +25,19 @@
...
 if TYPE_CHECKING:
-    from collections.abc import AsyncIterator, Iterable, MutableMapping
+    from collections.abc import AsyncIterator, Iterable, Mapping, MutableMapping
 
     from betty.assets import AssetRepository
     from betty.cache.file import BinaryFileCache
     from betty.user import User
 
 
+class UnknownTranslationError(Exception):
```
bartfeenstra: Add a factory method that sets a message that includes the locale for which a translation was requested

bartfeenstra: Should we instead return an UnknownLocaleError?

PR #2812, opened 2025-09-28
Add global HTTP rate limit functionality

### LINKED ISSUE #2793: Global rate limits
Rate limits are currently applied by the Wiki extension's `Client`. However, if we make it global through plugins and configuration, fetchers can apply limits automatically. Or can we use aiohttp middleware?

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2793


betty/http_client.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @abstractmethod
+    def match(self, request: ClientRequest) -> bool:
+        """
+        Whether this policy matches the given request.
+        """
+
+    @property
+    @abstractmethod
+    def rate_limiter(self) -> RateLimiter:
```
bartfeenstra: We only have a single rate limiter implementation. Would it make more sense to just return the maximum rate and let the middleware manage the actual limiters?

betty/app/__init__.py
```diff
@@ -215,16 +219,37 @@ async def localizers(self) -> LocalizerRepository:
...
         return LocalizerRepository(await self.translations)
 
+    @service
+    def http_rate_limits(self) -> PluginRepository[RateLimitDefinition]:
+        """
+        The rate limit plugin repository.
+
+        Read more about :doc:`/development/plugin/rate-limit`.
+        """
+        return EntryPointPluginRepository(RateLimitDefinition, "betty.rate_limit")
```
bartfeenstra: prefix

PR #2819, opened 2025-09-29
Introduce plugin definitions

### LINKED ISSUE #2816: Decouple plugins and their metadata
Benefits:
- Some plugins can be metadata without a class, also making it clearer when plugins do not require initialization 
- Plugins can be more easily defined outside of Python code

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2816


betty/plugin/__init__.py
```diff
@@ -46,68 +49,108 @@ class PluginError(Exception):
...
+class UserFacingPluginDefinition(UserFacing, PluginDefinition):
+    """
+    A definition of a plugin that is user-facing.
+    """
+
+    def __init__(
+        self,
+        *,
+        label: Localizable,
+        label_plural: Localizable,
```
bartfeenstra: Extract countability into a separate type?

betty/plugin/__init__.py
```diff
@@ -46,68 +49,108 @@ class PluginError(Exception):
...
 _PluginT = TypeVar("_PluginT", bound=Plugin)
 _PluginCoT = TypeVar("_PluginCoT", bound=Plugin, covariant=True)
 
 
+class ClassedPluginDefinition(Generic[_PluginT], PluginDefinition):
+    """
+    A definition of a plugin that is based around a class.
+    """
+
+    _cls: type[_PluginT]
```
bartfeenstra: Allow the class to be optionally set through the initializer.

betty/plugin/__init__.py
bartfeenstra: Or pluralization, specifically. Also find a solution for defining countable plurals.

betty/plugin/__init__.py
```diff
@@ -46,68 +49,108 @@ class PluginError(Exception):
...
+        """
+        The plugin class.
+        """
+        return self._cls
+
+    def __call__(self, cls: type[_PluginT]) -> type[_PluginT]:
+        """
+        Set the plugin's class.
+        """
+        self._cls = cls
```
bartfeenstra, line 324: Also set the definition on the plugin class.

betty/plugin/__init__.py
```diff
@@ -46,68 +49,108 @@ class PluginError(Exception):
...
     @classmethod
     def plugin_description(cls) -> Localizable | None:
         """
-        Get the human-readable long plugin description.
+        The human-readable long plugin description.
         """
         return None
 
 
+class Plugin(Generic[_PluginDefinitionT], ABC):
```
bartfeenstra: Use a protocol so we can reduce class hierarchies.

betty/plugin/__init__.py
bartfeenstra: Or we make it required and move decorator functionality to function-based decorators.

betty/ancestry/event_type/__init__.py
```diff
@@ -4,49 +4,63 @@
...
 
-    @final
-    @override
-    @classmethod
-    def plugin_type_cls(cls) -> type[Plugin]:
-        return EventType
+    def __init__(
+        self,
+        *,
+        is_start_of_life: bool = False,
```
bartfeenstra, line 51: Make these mutually exclusive. Add some kind of 'indicator' enum?

bartfeenstra: 'phase'?

bartfeenstra: But also there is before-during-after life, and start- and end-of-life indication. E.g. a funeral is both after-life and end-of-life.

bartfeenstra: In templates, however, we need boolean checks. The old code uses dedicated tests for start-of-life and end-of-life events.

betty/ancestry/event_type/event_types.py
```diff
@@ -46,308 +46,271 @@ def may_create(cls, person: Person, lifetime_threshold: int) -> bool:
         return True  # pragma: no cover
 
 
-class PreBirthEventType(EventType):
```
bartfeenstra, line 49: Check that these marker classes have been correctly refactored into definitions for all subclasses.

betty/plugin/__init__.py
```diff
@@ -45,143 +49,277 @@ class PluginError(Exception):
...
+
+_PluginDefinitionT = TypeVar("_PluginDefinitionT", bound=PluginDefinition)
+# @todo Do we need the covariant version of the typevar still?
+_PluginDefinitionCoT = TypeVar(
+    "_PluginDefinitionCoT", bound=PluginDefinition, covariant=True
+)
+
+
+@final
+class PluginTypeDefinition(Generic[_PluginDefinitionT]):
```
bartfeenstra: Does this need to be generic at all?

betty/plugin/__init__.py
```diff
@@ -45,143 +49,277 @@ class PluginError(Exception):
...
+    """
+    A plugin type definition.
+    """
+
+    def __init__(
+        self,
+        *,
+        id: MachineName,  # noqa A002
+        label: Localizable,
+        cls: type | None = None,
```
bartfeenstra: Can we create a `ClassedPluginTypeDefinition`?

betty/ancestry/citation.py
```diff
@@ -12,48 +12,39 @@
...
 
 
 @final
-class Citation(
-    ShorthandPluginBase,
-    HasDate,
-    HasFileReferences,
-    HasPrivacy,
-    HasLinks,
-    UserFacing,
```
bartfeenstra, line 48: Add a `user_facing=False` key to `EntityDefinition`

bartfeenstra: Do we still need `UserFacing` then?

bartfeenstra: We don't even need a key because our plugin definitions are user facing already

PR #2873, opened 2025-10-14
Cache all HTTP responses using aiohttp-client-cache

### LINKED ISSUE #2817: Add caching middleware to the HTTP client
Add middleware to the HTTP client that caches responses. That way we can remove this functionality from fetchers. 

## Proposals
- Take inspiration from https://github.com/requests-cache/aiohttp-client-cache. In fact, aiohttp-client-cache's `CachedSession` inherits from aiohttp's `ClientSession` so compatibility is guaranteed. We could just use its drop-in replacement and be done quickly.

### LINKED ISSUE #2818: Refactor fetchers to be for files only
Once https://github.com/bartfeenstra/betty/issues/2817 is merged, fetchers primary objective (to provide response caching) will have become obsolete. 

## To do
- Remove `Fetcher.fetch()`
- Refactor fetchers into file-only fetchers whose goal it is to make an HTTP response's file accessible as a file on disk.
  - Allow callers to provide an optional file destination path. If none is given, fetchers must choose a location themselves, such as in a binary file cache.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2817 and this fixes https://github.com/bartfeenstra/betty/issues/2818


betty/test_utils/conftest.py
```diff
@@ -140,3 +137,12 @@ async def page(context: BrowserContext) -> Page:
     page = await context.new_page()
     page.on("pageerror", do_raise)
     return page
+
+
+@pytest.fixture(autouse=True)
+def http_mock_responses(tmp_path: Path) -> Iterator[aioresponses]:
```
bartfeenstra: Temp path is not needed

PR #1903, opened 2024-08-29
Remove file operations from renderers

bartfeenstra:
## To do
- Remove `@todo`s

PR #2950, opened 2025-10-31
Remove file operations from renderers

bartfeenstra:
This obsoletes https://github.com/bartfeenstra/betty/pull/1903

### DIFF
--- a/betty/render.py
+++ b/betty/render.py
@@ -5,29 +5,34 @@ Provide the Render API.
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, ClassVar, final
+from collections.abc import Awaitable, Callable
+from contextlib import suppress
+from pathlib import Path
+from shutil import copy2
+from typing import TYPE_CHECKING, Any, ClassVar, TypeAlias, final
 
+from aiofiles.os import makedirs
 from typing_extensions import override
 
 from betty.locale.localizable import _
-from betty.plugin import (
-    ClassedPluginDefinition,
-    ClassedPluginTypeDefinition,
-)
+from betty.media_type import UnsupportedMediaType, match_extension, match_media_type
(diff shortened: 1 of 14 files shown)


betty/render.py
```diff
@@ -40,20 +42,66 @@ def media_types(self) -> Sequence[MediaType]:
...
+        *,
+        data: Mapping[str, Any] | None = None,
+        job_context: Context | None = None,
+        localizer: Localizer | None = None,
+    ) -> str:
+        """
+        Render content.
         """
 
+    def new_copy_function(
```
bartfeenstra: Use a different prefix as we use `new_` for creating new instances of classes.

betty/render.py
```diff
@@ -40,20 +42,66 @@ def media_types(self) -> Sequence[MediaType]:
...
+    ) -> str:
+        """
+        Render content.
         """
-        Render a single file.
 
-        :return: The file's new path, which may have been changed, e.g. a
-            renderer-specific extension may have been stripped from the end.
+    @abstractmethod
+    async def render_async(
```
bartfeenstra: Add a default implementation that calls `self.render()`

betty/render.py
```diff
@@ -40,20 +41,73 @@ def media_types(self) -> Sequence[MediaType]:
...
-    ) -> Path:
+    ) -> str:
+        """
+        Render content.
         """
-        Render a single file.
 
-        :return: The file's new path, which may have been changed, e.g. a
-            renderer-specific extension may have been stripped from the end.
+    def make_copy_function(
```
bartfeenstra: This function gives us the most headaches because it is a bit of a hacky integration between renderers (who no longer know about files), files, wrapped up in a single function just so we can reuse it 2-3 times.

bartfeenstra: It's also highly opinionated. Let's take this off `Renderer`, move it to a more appropriate module, and mark it `@internal`

PR #2958, opened 2025-11-02
Run CI tests using Python free threading

### LINKED ISSUE #2956: Run CI tests using Python's free threading
See https://docs.python.org/3/howto/free-threading-python.html.

We can depend on lxml 5 for Python 3.13 and below, see https://github.com/python/mypy/pull/20029#discussion_r2413378521

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2956

## To do
- allow Betty to be installed without Playwright so we avoid greenlet
- in a separate PR: update the test CI matrix to run a dedicated Playwright build. Also add a build command and pyproject.toml group to install Betty without Playwright.
- Add a pyproject.toml classifier
- add a test to ensure that if the current Python build support free threading, free threading is indeed enabled

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,6 +44,10 @@ jobs:
             os: windows-latest
             nodejs: 24.x
             python: '3.14'
+          - name: 'Free threading'
+            os: ubuntu-latest
+            nodejs: 22.x
+            python: '3.14t'
     timeout-minutes: 30
     steps:
       - name: Check out the source code


bartfeenstra:
CPython automatically enables the GIL because of the greenlet package, which is a Playwright dependency. Check **when** CPython determines it should reenable the GIL. If that's due to imports, we can simply avoid this by skipping Playwright tests. Except that purest loads all its plugins automatically...

PR #2962, opened 2025-11-02
Remove the dependency on Bash

### LINKED ISSUE #2894: Remove the dependency on Bash
Remove the dependency on Bash because it is no longer supported by macOS by default and we should really improve compatibility.

Refactor the commands using Python instead.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/2894


.github/workflows/test.yml
```diff
@@ -20,26 +20,27 @@ jobs:
...
-            python: '3.12'
-          - name: 'Python 3.11'
-            os: ubuntu-latest
-            nodejs: 22.x
-            python: '3.11'
-          - name: 'macOS'
-            os: macos-latest
-            nodejs: 22.x
-            python: '3.14'
+          # @todo Reenable the matrix
```
bartfeenstra: Fix this

PR #3066, opened 2025-11-11
Refactor the render API

### LINKED ISSUE #3031: Split renderers into media type converters
One input and one output media type. Allow a wildcard, needed for `Jinja2Renderer` which can render to any text media type, and for https://github.com/bartfeenstra/betty/issues/3026

Actually, refactor renderers into media type converters and keep Jinja2 as our only hard coded templating engine.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3031

## To do
- Split renderers up into templaters (type 2), converters (type 1), and utilities (type 3)
- We may just want to keep one templater (Jinja2), because we're unlikely to add another soon, while we do want multiple converters. That would let us simplify some of the Jinja2 code, and refactor renderers into converters.
- type 2 need a copy function 
- types 1 and 2 need logic. Type 3 never involve media type changes anyway

### DIFF
--- a/betty/render.py
+++ b/betty/render.py
@@ -98,13 +98,13 @@ class RendererDefinition(ClassedPluginDefinition[Renderer]):
 # @todo Type 1 is really the only conversion. Type 2 just resolves templating and leaves the rest intact.
 # @todo Type 3 may cause conflicts but that depends on whether individual renderers act on the same content.
 # @todo
+# @todo Types 1 and 2 are also pretty non-negotiable: given an input and desired output media type, we can either render
+# @todo or reliably error because we do not support that specific conversion.
 # @todo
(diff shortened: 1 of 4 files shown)


betty/render.py
```diff
@@ -30,20 +29,15 @@
...
 
-    @property
-    @abstractmethod
-    def media_types(self) -> Sequence[MediaType]:
-        """
-        The media types this renderer can render from.
-        """
-
     @abstractmethod
     async def render(
```
bartfeenstra, line 42: When rendering we must (also) pass on the required output type (as opposed to assume HTML which is what we have always been doing).

Alternatively we formally adopt the assumption that the output is always HTML. (But what about our json.j2 files then?)

bartfeenstra: Another argument in favor of separating the 3 renderer types out into separate APIs

PR #3084, opened 2025-11-13
Improve the content provider API DX

### LINKED ISSUE #3081: Improve content provider DX
## To do
- Put all templates in /content, and then /content/maps etc for extensions
- implement get_mutable_instances() on all configuration classes
- consider adding type aliases and helper methods for plugins who wrap other plugins and their configuration 
- Rename content.html.j2 and ensure test coverage for passing on resource contexts. Also split it up into a template that takes `ContentProvider` instances and one that takes `PluginInstanceConfiguration`

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3081


betty/content_provider/content_providers.py
```diff
@@ -62,7 +62,7 @@ class Jinja2TemplateContentProvider(
     Provides content by rendering a Jinja2 template.
     """
 
-    _template: str
+    _content_template_name: str
```
bartfeenstra: Actually, assemble the template name automatically from the plugin ID

bartfeenstra: Remove the attributes from subclasses

betty/project/extension/demo/content_provider.py
```diff
@@ -14,12 +14,12 @@
     label=Plain("Front page content (demo)"),
 )
 class _FrontPageContent(Jinja2TemplateContentProvider):
```
bartfeenstra: Remove the template and return it inline. 

Or instead, consider removing StaticTranslationsConfigurationAttr and allowing any Localizables to be set on configuration. We can then remove these plugins entirely. (do this in a separate PR)

bartfeenstra: Blocked by https://github.com/bartfeenstra/betty/issues/3088

betty/content_provider/content_providers.py
```diff
@@ -78,7 +76,9 @@ async def provide(self, *, resource: Context) -> str | None:
         await self._project.localizers
```
bartfeenstra: Remove this as it's unused

PR #3090, opened 2025-11-14
Make configuration accept any `Localizable`

### LINKED ISSUE #3088: Allow configuration to accept any `Localizable`
Allow configuration to accept any `Localizable` instead of just static translations. Use cases are hard coded configuration such as the Demo extension or user extensions, or content providers. 

We would need to handle serialization. For loading we can accept static translations only. For dumping, can we error if the localizable is not static translations?

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3088

### DIFF
--- a/betty/config/__init__.py
+++ b/betty/config/__init__.py
@@ -4,7 +4,7 @@ The Configuration API.
 
 from __future__ import annotations
 
-from typing import Any, Generic, Self, TypeVar
+from typing import Any, Generic, TypeVar
 
 from betty.mutability import Mutable
 from betty.serde.dump import Dumpable
@@ -15,13 +15,6 @@ class Configuration(Mutable, Loadable, Dumpable):
     """
     Any configuration object.
     """
-
-    def update(self, other: Self, /) -> None:
-        """
-        Update this configuration with the values from ``other``.
-        """
-        self.assert_mutable()
-        self.load(other.dump())
 
 
 _ConfigurationT = TypeVar("_ConfigurationT", bound=Configuration)
--- a/betty/content_provider/content_providers.py
+++ b/betty/content_provider/content_providers.py
@@ -8,14 +8,15 @@ from typing import TYPE_CHECKING, Any, Self
 
 from typing_extensions import override
 
-from betty.config import Configurable
+from betty.assertion import RequiredField, assert_record
+from betty.config import Configuration
+from betty.config.factory import ConfigurationDependentSelfFactory
 from betty.content_provider import ContentProvider, ContentProviderDefinition
-from betty.factory import IndependentFactory
 from betty.html import plain_text_to_html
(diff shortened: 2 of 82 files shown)


betty/locale/localizable/config.py
```diff
@@ -3,100 +3,36 @@
...
-
 
-@final
-class StaticTranslationsConfiguration(Configuration, StaticTranslations):
+def dump_if_static_translations(localizable: Localizable) -> Dump:
     """
-    Provide configuration for a :py:class:`betty.locale.localizable.Localizable`.
+    Dump static translations to configuration.
+
+    :raises betty.exception.HumanFacingException: raise if the localizable is not :py:class:`betty.locale.localizable.StaticTranslations`.
```
bartfeenstra: Add a new error class to the dump API that can we raised when something is not dumpable. Document that it can be raised by Dumpable.dump()

betty/content_provider/content_providers.py
```diff
@@ -22,6 +23,25 @@
 
     from betty.project import Project
     from betty.resource import Context
+    from betty.serde.dump import Dump
+
+
+class PlainTextConfiguration(Configuration):
```
bartfeenstra, line 32: Make this generic locale config because it is reusable. Extend `StaticTranslationsConfiguration`. Or just update that class.

betty/locale/localizable/config.py
```diff
@@ -2,101 +2,126 @@
...
 from betty.locale.localizable.assertion import assert_static_translations
 from betty.serde.dump import Dump
 
 _T = TypeVar("_T")
+_LocalizableT = TypeVar("_LocalizableT")
 
 
-@final
-class StaticTranslationsConfiguration(Configuration, StaticTranslations):
+class _StaticTranslationsPreferredLocalizableConfigurationAttr(Generic[_LocalizableT]):
```
bartfeenstra: Can we just call this "LocalizableConfiguration" because it's really the only config we offer that does localizable support. The details are not too relevant for the name as there is little need to distinguish it from other types (because don't exist).

betty/config/__init__.py
```diff
@@ -20,12 +21,17 @@ class Configuration(Mutable, Loadable, Dumpable):
     Any configuration object.
     """
 
-    def update(self, other: Self, /) -> None:
+
+class UpdatableConfiguration(Configuration):
```
bartfeenstra: Do we use this?

betty/plugin/config/__init__.py
```diff
@@ -220,11 +211,7 @@ def __init__(
     ):
         super().__init__()
```
bartfeenstra, line 213: Overload __init__() so we can properly type both cases:
- A resolvable plugin ID and a configuration dump
- A plugin class and a configuration instance

PR #3106, opened 2025-11-17
Allow plugins to depend on extensions

### LINKED ISSUE #3085: Allow plugins to depend on extensions
That way, if plugins need assets or services, we can prevent them from being used. 

Should we handle this inside plugin repositories? Do we want to add a new repository that decorates others and omits plugins with unmet dependencies, and emits warnings? 

## Blocked by
- https://github.com/bartfeenstra/betty/issues/3087

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3085

## To do
- Add `@requires_extension`

### DIFF
--- a/betty/plugin/__init__.py
+++ b/betty/plugin/__init__.py
@@ -316,6 +316,32 @@ _DependentPluginDefinitionT = TypeVar(
 )
 
 
+class ExtensionDependentPluginDefinition(PluginDefinition):
+    """
+    A definition of a plugin that can declare its dependency on extensions.
+    """
+
+    def __init__(
+        self,
+        *,
+        depends_on_extensions: Set[PluginIdentifier] | None = None,
+        **kwargs: Any,
+    ):
+        super().__init__(**kwargs)
+        self._depends_on_extensions = (
+            set()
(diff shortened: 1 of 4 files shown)


betty/plugin/__init__.py
```diff
@@ -316,6 +316,32 @@ def depends_on(self) -> Set[MachineName]:
...
 
+class ExtensionDependentPluginDefinition(PluginDefinition):
+    """
+    A definition of a plugin that can declare its dependency on extensions.
+    """
+
+    def __init__(
+        self,
+        *,
+        depends_on_extensions: Set[PluginIdentifier] | None = None,
```
bartfeenstra, line 327: Rename to `requires_extension`

betty/plugin/__init__.py
```diff
@@ -316,6 +316,32 @@ def depends_on(self) -> Set[MachineName]:
 )
 
 
+class ExtensionDependentPluginDefinition(PluginDefinition):
```
bartfeenstra, line 319: Now that we want to introduce `@required_extension`, we no longer need to change anything about definitions.

PR #3107, opened 2025-11-17
Allow plugins to express system requirements

### LINKED ISSUE #3086: Allow plugins to declare system requirements
Do we want to add a new repository that decorates others and omits plugins with unmet requirements, and emits warnings? 

## Blocked by
- https://github.com/bartfeenstra/betty/issues/3087

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3086

## To do
- Refactor all plugin repository services to be marked `@internal` or just make them all private to be really, really clear.
- Make `PluginRepositoryProvider` internal
- Refactor all code calling to plugin repository services to use plugin type definitions instead
- Refactor the original extension requirement handling to leverage this new solution

### DIFF
--- a/betty/plugin/__init__.py
+++ b/betty/plugin/__init__.py
@@ -505,11 +505,11 @@ class PluginRepositoryDefinition(Generic[_PluginDefinitionT], ABC):
         """
         Get the repository for this plugin type.
         """
-        from betty.plugin.requirement import new_requirement_met_plugin_repository
+        from betty.plugin.requirement import PluginRequirementRepository
 
         repository = await self._get(service_level)
         if not unavailable:
-            repository = await new_requirement_met_plugin_repository(
+            repository = await PluginRequirementRepository.new(
                 repository, service_level
             )
         return repository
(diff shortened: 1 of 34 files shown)


betty/plugin/__init__.py
```diff
@@ -462,13 +499,26 @@ class PluginRepositoryDefinition(Generic[_PluginDefinitionT], ABC):
...
+        self, service_level: ServiceLevel = None, *, unavailable: bool = False
     ) -> PluginRepository[_PluginDefinitionT]:
         """
         Get the repository for this plugin type.
         """
+        from betty.plugin.requirement import PluginRequirementRepository
+
+        repository = await self._get(service_level)
+        if not unavailable:
+            repository = await PluginRequirementRepository.new(
```
bartfeenstra, line 513: Requirements can be expensive (I/O). How do we cache this? We can cache global repos here. What do to for apps and projects? They have service-based caching...

bartfeenstra: If we move this logic to App and Project, they can handle all the caching.

bartfeenstra: We'd have three entry points instead of one (a third for no service level, a global function somewhere), but have all the caching benefits

PR #3234, opened 2025-11-26
Use Babel's `Locale` everywhere to reduce data conversions

### LINKED ISSUE #3230: Improve locale performance
Profiling shows a lot of time is spent parsing locale strings into Babel `Locale` objects. Can we refactor our internals to use `Locale` objects everywhere and only parse strings for things like entities, configuration, etc.?

If I recall correctly, the original reason for keeping locales as strings and only getting Babel objects when needed, was to support non-system locales as much as possible. However, we never actually bothered to guarantee or even test for that. Maybe it is time to formally let go of this assumption and use Babel `Locale` objects in our APIs directly.

There is also the issue of 'meta' locales that don't refer to a specific locale: `zxx`, `und`, `mis`, and `mul`.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3230

### DIFF
--- a/betty/locale/__init__.py
+++ b/betty/locale/__init__.py
@@ -7,72 +7,52 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, TypeAlias
 
 from babel import Locale
-from babel import negotiate_locale as babel_negotiate_locale
 from babel.core import UnknownLocaleError
-from langcodes import Language
 
 import betty.dirs
 
 if TYPE_CHECKING:
     from collections.abc import Sequence
 
+
 _LOCALE_DIRECTORY_PATH = betty.dirs.ASSETS_DIRECTORY_PATH / "locale"
 
-NO_LINGUISTIC_CONTENT = "zxx"
-UNDETERMINED_LOCALE = "und"
-UNCODED_LOCALE = "mis"
-MULTIPLE_LOCALES = "mul"
-SPECIAL_LOCALES = (
-    NO_LINGUISTIC_CONTENT,
-    UNDETERMINED_LOCALE,
-    UNCODED_LOCALE,
-    MULTIPLE_LOCALES,
-)
 
-DEFAULT_LOCALE = "en-US"
+DEFAULT_LOCALE = Locale("en", "US")
+"""
+Betty's default locale (US English).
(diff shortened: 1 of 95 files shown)


betty/locale/__init__.py
```diff
@@ -67,12 +61,11 @@ def to_babel_identifier(locale: LocaleLike, /) -> str:
     )
 
 
-def to_locale(locale: LocaleLike, /) -> str:
+# @todo Ensure we only really use this to generate public data, and we use Babel's format internally.
+def to_language_tag(locale: Locale, /) -> str:
```
bartfeenstra: This should take `None` as well and convert that to `und`.

betty/locale/__init__.py
```diff
@@ -84,18 +79,14 @@ def to_locale(locale: LocaleLike, /) -> str:
     )
 
 
-LocaleLike: TypeAlias = str | Locale
-
-
-def get_data(locale: LocaleLike, /) -> Locale:
+# @todo Ensure we only really use this to generate public data, and we use Babel's format internally.
+def from_language_tag(locale: str, /) -> Locale:
```
bartfeenstra, line 67: We should not need as IETF language tags are for web publishing only.

betty/locale/__init__.py
```diff
@@ -128,21 +119,23 @@ def negotiate_locale(
     """
     Negotiate the preferred locale from a sequence.
     """
-    if isinstance(preferred_locales, str | Locale):
```
bartfeenstra, line 131: Can we remove this entire function and just use Babel's directly?

betty/locale/__init__.py
```diff
@@ -114,51 +112,37 @@ def get_display_name(
...
     /,
 ) -> Locale | None:
     """
     Negotiate the preferred locale from a sequence.
     """
-    if isinstance(preferred_locales, str | Locale):
+    if isinstance(preferred_locales, Locale):
         preferred_locales = [preferred_locales]
-    return _negotiate_locale(
-        [to_babel_identifier(locale) for locale in preferred_locales],
```
bartfeenstra, line 134: Can't we just use str() here?

PR #3271, opened 2025-11-30
Simplify `HumanFacingExceptionGroup`

bartfeenstra:


betty/config/file.py
```diff
@@ -36,19 +35,13 @@ async def assert_configuration_file(
     }
 
     def _assert(configuration_file_path: Path) -> _ConfigurationT:
-        with (
-            HumanFacingExceptionGroup().assert_valid() as errors,
-            # Change the working directory to allow relative paths to be resolved
-            # against the configuration file's directory path.
-            chdir(configuration_file_path.parent),
```
bartfeenstra, line 43: We accidentally removed the `chdir()` call.

betty/exception.py
```diff
@@ -161,30 +177,24 @@ def append(self, *errors: HumanFacingException) -> None:
...
-    def _copy(self) -> Self:
-        return type(self)()
+    def within_context(self, *contexts: Context) -> None:
+        self._contexts.extend(contexts)
+        for error in self._errors:
+            error.within_context(*contexts)
 
     @contextmanager
-    def catch(self, *contexts: Context) -> Iterator[HumanFacingExceptionGroup]:
+    def catch(self, *contexts: Context) -> Iterator[None]:
```
bartfeenstra: Rename to `absorb`

betty/config/file.py
bartfeenstra: Did we even have test coverage for this? If not, add it.

PR #3281, opened 2025-12-05
Make `Loadable.load()` a class method

### LINKED ISSUE #3277: Make `Configuration.load()` a class method
We currently have to create configuration with fake values all the time. That's prone to bugs. A class method would avoid all that.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3277


betty/assertion.py
```diff
@@ -113,6 +112,7 @@ class Field(Generic[_AssertionValueT, _AssertionReturnT]):
 
     name: str
     assertion: Assertion[_AssertionValueT, _AssertionReturnT] | None = None
+    as_name: str | None = None
```
bartfeenstra, line 115: Add test coverage

PR #3299, opened 2025-12-09
Add factories back to plugin repositories

### LINKED ISSUE #3125: Add factories back to plugin repositories
Now that plugin repositories are/will be provided only by app, project, or the global service level, we can reliably inject factories. Allow the factory methods to take optional configuration, and move the relevant code from `PluginInstanceConfiguration` to repos.

## Blocked by
- https://github.com/bartfeenstra/betty/issues/3216

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3125

**ACTUALLY...**
We're having lots of trouble getting everything to be type-safe because Python does not support higher kinded types. That's an argument for keeping plugin classes out of plugin repos so they are only generic over definitions. Additionally, all this factory stuff mostly isn't specific to plugins: it's specific to configurable targets. Acknowledge that by moving this functionality next to existing factory functionality rather than in a more specific API such as the plugin API. Or add it to `betty.config.factory`.

## Blocked by
- https://github.com/bartfeenstra/betty/pull/3300

### DIFF
--- a/betty/plugin/repository/__init__.py
+++ b/betty/plugin/repository/__init__.py
@@ -9,16 +9,23 @@ from typing import TYPE_CHECKING, Generic
 
 from typing_extensions import TypeVar
 
+from betty.config import Configurable, Configuration
+from betty.config.factory import ConfigurationDependentSelfFactory
+from betty.exception import HumanFacingException
+from betty.importlib import fully_qualified_name
 from betty.json.schema import Enum
+from betty.locale.localizable.gettext import _
 from betty.locale.localizer import DEFAULT_LOCALIZER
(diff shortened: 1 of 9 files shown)


betty/plugin/repository/__init__.py
```diff
@@ -76,3 +86,38 @@ def plugin_id_schema(self) -> Enum:
                 description=f"A {label} plugin ID",
             )
         return self._plugin_id_schema
+
+    async def new_target(
```
bartfeenstra, line 90: `new_instance`

bartfeenstra: Same for the config

PR #3300, opened 2025-12-09
Allow type-safe access from plugins to plugin definitions to plugin type definitions

### LINKED ISSUE #3292: Allow type-safe access from plugins to plugin definitions to plugin type definitions
That way we can type its return value using generics

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3292

### DIFF
--- a/betty/plugin/__init__.py
+++ b/betty/plugin/__init__.py
@@ -10,11 +10,13 @@ Read more at :doc:`/development/plugin`.
 from __future__ import annotations
 
 from contextlib import contextmanager
+from functools import update_wrapper
 from importlib import metadata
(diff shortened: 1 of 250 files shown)


betty/plugin/__init__.py
```diff
@@ -84,16 +78,16 @@ def cls(self) -> builtins.type[_PluginCoT]:
         assert self._cls is not None
         return self._cls
 
-    def __call__(self, cls: builtins.type[_PluginCoT]) -> builtins.type[_PluginCoT]:
+    def __call__(self, cls: builtins.type[_T]) -> builtins.type[_T]:
```
bartfeenstra: Extract the decoration functionality of both definition classes into a new private superclass.

PR #3465, opened 2026-01-06
Simplify the `ProjectConfiguration` dump

### LINKED ISSUE #3461: Simplify the `ProjectConfiguration` dump
See https://betty.readthedocs.io/0.5.x/betty.project.config/#betty.project.config.ProjectConfiguration. Minimal dumps must not include `None`/`null` and strip empty collections.

### LINKED ISSUE #3466: Add `ConfigurationTestBase`
## Definition of Done
- For each sample, test that when it is dumped, that dump is equal to that same dump being loaded and dumped again. This ensures consistency and small dumps. 
- Potentially allow samples to be marked not for documentation, to indicate they are relevant for testing but don't add much for compared to the other samples when read by humans. 
- Ensure that all samples include unique configuration, by comparing the configuration objects against each other, but also their dumps.
- For the configuration in each sample, assert it is equal to itself.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3461

### DIFF
--- a/betty/project/config.py
+++ b/betty/project/config.py
@@ -40,7 +40,6 @@ from betty.exception import (
     reraise_within_context,
 )
 from betty.license import License, LicenseDefinition
-from betty.license.licenses import AllRightsReserved
 from betty.locale import DEFAULT_LOCALE, LocaleLike, ensure_locale, to_language_tag
 from betty.locale.localizable.assertion import assert_load_localizable
 from betty.locale.localizable.attr import (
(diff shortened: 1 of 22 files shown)


betty/project/config.py
```diff
@@ -847,17 +847,19 @@ def __init__(
         self._entity_types = (
             EntityTypeConfigurationMapping() if entity_types is None else entity_types
         )
-        self.copyright_notice = copyright_notice or PluginInstanceConfiguration[
+        self._default_copyright_notice = PluginInstanceConfiguration[
```
bartfeenstra: These defaults won't work in case they are altered after initialization. We should probably set them, and when dumping do a deep comparison to see if they are the default.

bartfeenstra: Use `__eq__()`

Issue #3217, opened 2025-11-25
Automatically dump linked data for localizable attributes again


bartfeenstra:
We removed this sometime in the spring or summer of 2025 when we got rid of the previous generation of localizable attributes. Bring this back.  See https://github.com/bartfeenstra/betty/pull/1946

## To do
- Remove the current dumping code from all entities 

## Blocked by
- https://github.com/bartfeenstra/betty/issues/3449

bartfeenstra:
Closing in favor of https://github.com/bartfeenstra/betty/issues/4164

PR #3481, opened 2026-01-08
Add data definitions

### LINKED ISSUE #3288: Propagate configuration validation
## Proposals
- Allow all configuration to declare contained configuration by specifying data context selectors. 
- Update selector classes with a method to extract the data they represent from an object. 
- Make `Configuration.mutables` return all contained configuration based on these selectors. 
- Make `Configuration.validator` propagate to contained configuration 

## Blocked by
- https://github.com/bartfeenstra/betty/issues/3449

### LINKED ISSUE #3424: Document configuration objects
Move configuration documentation out of regular `*.rst `pages and move it onto the respective configuration classes. Limit examples to just YAML. 

Possibly add a test that ensures all public and non-abstract configuration classes have docstrings that included code examples.

### LINKED ISSUE #3449: Data definition API
This would improve reusability, DX, and build foundations for building config UIs automatically.

## Proposals 
- Use descriptors that implement the new API
  - attr name
  - optional/required (can we make attrs required using a decorating attr? That would save some duplication)
- These descriptors also handle serialization (serde and/or linked data?)
- That way, `ConfigurationMapping` and `ConfigurationSequence` can implement serde on behalf of all subclasses. 
- split data selectors into indicators and selectors (which are indicators). An indicator may be "any key" or "any index" or a path or a URL, and not be usable as a concrete selector.
- support read-only fields (for final or computed fields)
- types
  - simple types (scalars: int, float, str, bool, None)
  - complex types (they expose contained items along with data descriptors)
    - records (objects with descriptor attributes)
    - collections
      - mappings
      - sequences 
- How to support unions? We could add dedicated support for `None`. Would we need any other unions?
- Use the declarations to implement `Configuration.__eq__()` once on behalf of all subclasses. Make this implementation final.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3449

## Follow-ups
- expand the has_data Sphinx directive to recursively render data elements. Do this for simple as well as aggregate root data.
- Allow mixin classes for **records** to declare their own field that can automatically be found by the subclass that is the actual record data.
- For a project's extension configuration, replace the `enable()` method with the ability to add a plain resolvable plugin identifier that is expanded to a `PluginInstanceConfiguration` internally. Do this for locale configurations as well. Add this as resolvers on `KeyedCollection`. Also add `KeyedPluginInstanceConfigurationCollection`.
- Refactor `PluginInstanceConfigurationSequence*` to be regular sequences with value resolvers. Add a `ResolvingMutableSequence` which decorates another sequence but resolves values before passing them on.
- Keep `LocaleConfigurationMapping`, but rename it and refactor it onto `KeyedCollection` to add the default value behavior.
- Model plugins (entities, event types, etc) are also `HasData`. Should plugin and plugin type definitions subclass `DataDefinition`? We can override `data()` automatically like we do with `plugin()`. We should probably make the relevant plugin classes subclass `HasData` now and provide a default implementation without element. We can then gradually expand in the future without breaking anything.
- Fix `@todo`s
- Remove configuration collections 
- Remove `Configuration`
- Remove all `config.py` files
- Remove the Sphinx directives for configuration
- Generate samples automatically.
- Improve samples. Add `Samples` and put the functionality to get minimal and full samples on there. Merge `minimal` and `full` into a new `SampleSize` where the items have numerical values that we can sort on. Add a third option 'intermediate` which is the default for any sample.
- https://github.com/bartfeenstra/betty/issues/3217

### DIFF
--- a/betty/collections.py
+++ b/betty/collections.py
@@ -1,0 +1,136 @@
+"""
+Collection data tpes.
+"""
+
+from collections.abc import Callable, Collection, Iterable, Iterator, MutableSequence
+from typing import Generic, TypeVar, final, overload
+
+from typing_extensions import override
+
+from betty.functools import passthrough
+
+_KeyT = TypeVar("_KeyT")
+_ValueT = TypeVar("_ValueT")
+_ResolvableKeyT = TypeVar("_ResolvableKeyT")
+_ResolvableValueT = TypeVar("_ResolvableValueT")
+
+
+class KeyedCollection(
+    Collection[_ValueT], Generic[_KeyT, _ValueT, _ResolvableKeyT, _ResolvableValueT]
+):
+    """
+    A collection of values, automatically keyed.
+    """
+
+    def __init__(
+        self,
+        values: Iterable[_ResolvableValueT],
+        *,
+        key: Callable[[_ValueT], _KeyT],
+        key_resolver: Callable[[_ResolvableKeyT | _KeyT], _KeyT] = passthrough,
+        value_resolver: Callable[[_ResolvableValueT | _KeyT], _ValueT] = passthrough,
+    ):
+        self._values = {}
+        for value in values:
+            value = value_resolver(value)
+            self._values[key(value)] = value
+        self._key = key
+        self._key_resolver = key_resolver
+        self._value_resolver = value_resolver
+
+    @override
+    def __len__(self) -> int:
+        return self._values.__len__()
+
+    @override
+    def __iter__(self) -> Iterator[_ValueT]:
+        yield from self._values.values()
+
+    @override
+    def __contains__(self, key: object) -> bool:
+        try:
+            return (
+                self._key_resolver(
+                    key,  # ty:ignore[invalid-argument-type]
+                )
+                in self._values
+            )
+        except Exception:
+            return False
+
+    def __getitem__(self, key: _ResolvableKeyT) -> _ValueT:
+        return self._values[self._key_resolver(key)]
+
+    def __delitem__(self, key: _ResolvableKeyT) -> None:
+        del self._values[self._key_resolver(key)]
+
+    def add(self, value: _ResolvableValueT) -> None:
+        """
+        Add a value to the collection.
+        """
+        value: _ValueT = self._value_resolver(value)
+        self._values[self._key(value)] = value
+
+
+@final
+class ResolvingMutableSequence(
+    MutableSequence[_ValueT], Generic[_ValueT, _ResolvableValueT]
+):
+    """
+    A sequence of values.
+    """
+
+    def __init__(
+        self,
+        decorated: MutableSequence[_ValueT],
+        resolver: Callable[[_ResolvableValueT | _ValueT], _ValueT],
+        /,
+    ):
+        self._decorated = decorated
+        self._resolver = resolver
+
+    @override
+    def insert(self, index: int, value: _ResolvableValueT | _ValueT) -> None:
+        self._decorated.insert(index, self._resolver(value))
+
+    @overload
+    def __getitem__(self, index: int) -> _ValueT:
+        pass
+
+    @overload
+    def __getitem__(self, index: slice) -> MutableSequence[_ValueT]:
+        pass
+
+    def __getitem__(self, index):
+        return self._decorated[index]
+
+    @overload
+    def __setitem__(self, index: int, value: _ResolvableValueT | _ValueT) -> None: ...
+
+    @overload
+    def __setitem__(
+        self, index: slice, value: Iterable[_ResolvableValueT | _ValueT]
+    ) -> None: ...
+
+    def __setitem__(self, index, value):
+        if isinstance(index, int):
+            self._decorated[index] = self._resolver(value)
+        else:
+            self._decorated[index] = map(self._resolver, value)
+
+    def __delitem__(self, index: int | slice) -> None:
+        del self._decorated[index]
+
+    def __len__(self):
+        return len(self._decorated)
+
+    def __contains__(self, value: object) -> bool:
+        try:
+            return (
+                self._resolver(
+                    value,  # ty:ignore[invalid-argument-type]
+                )
+                in self._decorated
+            )
+        except Exception:
+            return False
--- a/betty/data/__init__.py
+++ b/betty/data/__init__.py
@@ -1,3 +1,251 @@
 """
-Data management.
+Describe, access, and manipulate arbitrary data.
 """
+
+from __future__ import annotations
+
+from functools import update_wrapper
+from typing import TYPE_CHECKING, Any, Generic, Self
+
+from typing_extensions import TypeVar
+
+from betty.importlib import fully_qualified_name
+from betty.locale.localizable.ensure import ensure_localizable
+from betty.portable import Portable, PortableData, PortablePorter, Porter
+from betty.portable.error import NotPortable
+from betty.service.hydrate import Hydratable
+
+if TYPE_CHECKING:
+    from collections.abc import Callable, Iterable
+
+    from ty_extensions import Intersection
+
+    from betty.locale.localizable import Localizable, LocalizableLike
+    from betty.service.level import ServiceLevel
+
+_DataT = TypeVar("_DataT", default=Any)
+
+
+class DataDefinition(Generic[_DataT]):
+    """
+    A data definition.
+    """
+
+    def __init__(
+        self,
+        *,
+        cls: type[_DataT] | None = None,
+        label: LocalizableLike,
+        description: LocalizableLike | None = None,
+        porter: Porter[_DataT] | None = None,
+        fallback_porter: Porter[_DataT] | None = None,
+        samples: Iterable[Callable[[], Sample[_DataT]]] | None = None,
+        empty: Callable[[_DataT], bool] | None = None,
+    ):
+        self._cls: type[_DataT] | None = None
+        self._label = ensure_localizable(label)
+        self._description = (
+            None if description is None else ensure_localizable(description)
+        )
+        self._porter = porter
+        self._fallback_porter = fallback_porter
+        self._samples = () if samples is None else list(samples)
+        self._empty = empty
+        if cls is not None:
+            self._cls = cls
+
+    @property
+    def cls(self) -> type[_DataT]:
+        """
+        The data's Python type.
+        """
+        if self._cls is None:
+            raise ValueError("This definition was not yet used to decorate a class.")
+        assert self._cls is not None
+        return self._cls
+
+    @property
+    def porter(self) -> Porter[_DataT]:
+        """
+        The porter for the data.
+        """
+        if self._porter is None:
+            if issubclass(self.cls, Portable):
+                self._porter = PortablePorter(self.cls)
+            elif self._fallback_porter is not None:
+                self._porter = self._fallback_porter
+
+            else:
+                raise NotPortable(
+                    f"This definition does not have a porter. Either make the data class {fully_qualified_name(self.cls)} subclass {fully_qualified_name(Portable)}, or provide a porter when initializing the definition."
+                )
+        assert self._porter is not None
+        return self._porter
+
+    def __call__(self, cls: type[_DataT]) -> type[_DataT]:
+        """
+        Decorate a data class.
+
+        :raises ValueError: Raised if the definition was already used to decorate a class.
+        """
+        if self._cls is not None:
+            raise ValueError("This definition was already used to decorate a class.")
+        if not issubclass(cls, HasData):
+            raise ValueError(
+                f"Can only decorate classes that subclass {fully_qualified_name(HasData)}."
+            )
+        assert self._cls is None
+        cls.data = staticmethod(update_wrapper(lambda: self, cls.data))  # ty:ignore[invalid-assignment]
+        self._cls = cls
+        return cls
+
+    @property
+    def label(self) -> Localizable:
+        """
+        The human-readable data label.
+        """
+        return self._label
+
+    @property
+    def description(self) -> Localizable | None:
+        """
+        The human-readable long data description.
+        """
+        return self._description
+
+    @property
+    def samples(self) -> Iterable[Sample[_DataT]]:
+        """
+        Any samples for this data.
+        """
+        for sample in self._samples:
+            yield sample()
+
+    def load(self, portable: PortableData, /) -> _DataT:
+        """
+        Create a new data instance from portable.
+
+        :raises betty.exception.HumanFacingException: Raised if the portable data is invalid.
+        """
+        return self.porter.load(portable)
+
+    def dump(self, data: _DataT, /) -> PortableData:
+        """
+        Dump the data to portable data.
+        """
+        return self.porter.dump(data)
+
+    async def hydrate(self, data: _DataT, services: ServiceLevel, /) -> None:
+        """
+        Hydrate the data.
+
+        Hydration allows data definitions to require a :py:type:`betty.service.level.ServiceLevel` to perform tasks
+        such as validation or enhancing the data using information or functionality from the service level.
+        """
+        from betty.config import Configuration
+
+        if isinstance(data, Hydratable):
+            await data.hydrate(services)
+        if isinstance(data, Configuration):
+            assert services is not None
+            validator = data.validator
+            if validator is not None:
+                await services.new_target(validator)
+
+    def empty(self, data: _DataT, /) -> bool:
+        """
+        Check if the data can be considered 'empty'.
+        """
+        if self._empty is None:
+            return False
+        return self._empty(data)
+
+
+_DataDefinitionT = TypeVar(
+    "_DataDefinitionT", bound=DataDefinition, default=DataDefinition, covariant=True
+)
+
+
+class HasData(Generic[_DataDefinitionT]):
+    """
+    A class that defines data for its instances.
+    """
+
+    @classmethod
+    def data(cls) -> Intersection[_DataDefinitionT, DataDefinition[Self]]:
+        """
+        Define the data for instances of this class.
+        """
+        raise NotImplementedError(
+            f"{fully_qualified_name(cls)} was not decorated with a {fully_qualified_name(DataDefinition)} subclass."
+        )
+
+    def __eq__(self, other: Any) -> bool:
+        if type(self) is not type(other):
+            return NotImplemented
+        data = type(self).data()
+        return data.dump(self) == data.dump(other)
+
+
+class Sample(Generic[_DataT]):
+    """
+    A data sample.
+
+    Samples are useful for generating documentation and tests.
+    """
+
+    def __init__(
+        self,
+        data: _DataT,
+        *,
+        label: LocalizableLike,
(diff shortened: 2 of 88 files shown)


betty/data.py
```diff
@@ -123,3 +133,117 @@ def __init__(self, path: pathlib.Path):
...
+
+    @property
+    def description(self) -> Localizable | None:
+        """
+        The human-readable long data description.
+        """
+        return self._description
+
+
+class ComplexData(Data):
```
bartfeenstra: How to handle simple data types? We want to support built-ins so we cannot use base classes for them. We can hardcode their definitions without a class, bit we also want to support fields such as machine name, which is just a string but comes with its own metadata and assertions.

bartfeenstra: We may need a second definition type to be able to work without classes, and be specifically for simple value types only.

betty/data.py
```diff
@@ -123,3 +133,117 @@ def __init__(self, path: pathlib.Path):
...
+    """
+
+
+class DataRecord(ComplexData):
+    """
+    A record with attributes.
+    """
+
+
+class DataCollection(ComplexData):
```
bartfeenstra: We want collections to be able to be defined on the fly, which means we would need to create definitions on the fly without having to write new classes. Can we make `ConfigurationCollection.Definition` a class var that contains the definition class for whatever collection class it is on? It can then take whatever it needs to create a definition specifically for its collection type .

betty/data/__init__.py
```diff
@@ -1,3 +1,176 @@
...
+        :raises ValueError: Raised if the definition was already used to decorate a class.
+        """
+        if self._cls is not None:
+            raise ValueError("This definition was already used to decorate a class.")
+        assert self._cls is None
+        self._cls = cls
+        return cls
+
+
+class DataCollection:
```
bartfeenstra: Do we need these still? Any serde assertions go in the definitions. Anything left like "this sequence must have at least one item" is up to the sequence itself. If somebody wants to use a plain list, they can. If somebody wants to use a sequence with extra behavior, they can do that too.

betty/data/__init__.py
```diff
@@ -1,3 +1,148 @@
...
+        description: LocalizableLike | None = None,
+    ):
+        self._value_cls = value_cls
+        self._label = ensure_localizable(label)
+        self._description = (
+            None if description is None else ensure_localizable(description)
+        )
+
+    @property
+    def value_cls(self) -> type[_ValueT]:
```
bartfeenstra: Just `cls`

betty/data/__init__.py
```diff
@@ -1,3 +1,148 @@
...
+from typing_extensions import TypeVar
+
+from betty.locale.localizable.ensure import ensure_localizable
+
+if TYPE_CHECKING:
+    from betty.locale.localizable import Localizable, LocalizableLike
+
+
+_ValueT = TypeVar("_ValueT", default=Any)
+_DataT = TypeVar("_DataT", default=Any)
```
bartfeenstra, line 26: Remove

betty/data/__init__.py
```diff
@@ -1,3 +1,148 @@
...
+
+from typing_extensions import TypeVar
+
+from betty.locale.localizable.ensure import ensure_localizable
+
+if TYPE_CHECKING:
+    from betty.locale.localizable import Localizable, LocalizableLike
+
+
+_ValueT = TypeVar("_ValueT", default=Any)
```
bartfeenstra: _T

betty/data/__init__.py
```diff
@@ -1,3 +1,148 @@
...
+
+    @property
+    def description(self) -> Localizable | None:
+        """
+        The human-readable long data description.
+        """
+        return self._description
+
+
+class ComplexDataDefinition(DataDefinition[_ValueT], Generic[_ValueT, _DataT]):
```
bartfeenstra: Add a method to get all contained data definitions.

bartfeenstra: Add mutability? We can't use `Mutable` (we should remove that later on) as the definitions must act on instances.

betty/data.py
bartfeenstra: When taking a data class for a definition, can we allow assertions? Then for machine names we can set the type to `assert_str() | str` and be done.

betty/data/__init__.py
```diff
@@ -1,3 +1,325 @@
...
+    A data definition.
+    """
+
+    def __init__(
+        self,
+        *,
+        cls: type[_DataT],
+        label: LocalizableLike,
+        description: LocalizableLike | None = None,
+        assertion: Assertion[_DataSetT, _DataT] | None = None,
```
bartfeenstra: `transformer`

betty/data/__init__.py
```diff
@@ -1,3 +1,325 @@
...
+    DataDefinition[_DataT, _DataSetT],
+    Generic[_DataT, _DataSetT, _DataItemT, _DataItemSetT, _IndicatorT, _SelectorT],
+):
+    """
+    A complex data definition.
+    """
+
+    @property
+    @abstractmethod
+    def data(self) -> Iterable[tuple[_IndicatorT, DataDefinition[Any, Any]]]:
```
bartfeenstra: Each item must also have a label and description to indicate the meaning of the item on the parent record data. They should be optional. If none are given, fallback to the element data definition label and description

betty/data/__init__.py
```diff
@@ -1,3 +1,325 @@
...
+
+    @property
+    def description(self) -> Localizable | None:
+        """
+        The human-readable long data description.
+        """
+        return self._description
+
+
+class ComplexDataDefinition(
```
bartfeenstra: AggregateDataDefinition

betty/data/__init__.py
bartfeenstra: `elements`

betty/data/__init__.py
```diff
@@ -1,3 +1,325 @@
...
+    def delete(self, data: _DataT, selector: _SelectorT, /) -> None:
+        """
+        Delete the value for contained data.
+        """
+
+
+@final
+class RecordDefinition(ComplexDataDefinition[_DataT, _DataSetT, Any, Any, Attr, Attr]):
+    """
+    A record data definition.
```
bartfeenstra: Explain what this means

betty/data/__init__.py
```diff
@@ -1,3 +1,325 @@
...
+        """
+        return self._description
+
+
+class ComplexDataDefinition(
+    DataDefinition[_DataT, _DataSetT],
+    Generic[_DataT, _DataSetT, _DataItemT, _DataItemSetT, _IndicatorT, _SelectorT],
+):
+    """
+    A complex data definition.
```
bartfeenstra: Explain what this means

betty/data/__init__.py
```diff
@@ -1,3 +1,325 @@
...
+    """
+
+    @property
+    @abstractmethod
+    def data(self) -> Iterable[tuple[_IndicatorT, DataDefinition[Any, Any]]]:
+        """
+        Get the data definitions contained by this one.
```
bartfeenstra: No verb, it's a property

betty/data/__init__.py
bartfeenstra: Add a transform() method. Call it in the aggregate types

betty/data/__init__.py
```diff
@@ -1,3 +1,325 @@
...
+class MappingDefinition(
+    ComplexDataDefinition[
+        _MutableMappingGetT, _MutableMappingSetT, _DataItemT, _DataItemSetT, AnyKey, Key
+    ]
+):
+    """
+    A key-value mapping data definition.
+    """
+
+    def __init__(
```
bartfeenstra, line 197: Extract this as well as length constraints into CollectionDefinition

bartfeenstra: Also the indicator as an init arg?

betty/data/__init__.py
```diff
@@ -1,3 +1,392 @@
...
+            label=label,
+            description=description,
+            elements={
+                Attr(selector): element for selector, element in elements.items()
+            },
+            factory=factory,
+        )
+
+
+class TypedDictDefinition(RecordDefinition[_MutableMappingT, Key, Key]):
```
bartfeenstra: Reconsider this name as it's not specific to dicts

bartfeenstra: Or just make it specific to dicts

betty/collection.py
```diff
@@ -0,0 +1,124 @@
...
+    def __delitem__(self, item: slice) -> None:
+        pass
+
+    @override
+    def __delitem__(self, item):
+        self._decorated.__delitem__(item)
+
+
+@final
+class KeyedSequence(MutableSequenceDecorator[_ValueT]):
```
bartfeenstra: This is wrong. Outwardly, this must behave in a set-like manner (it cannot fully be a set because we cannot implement the necessary comparison methods in a way that makes sense) plus extra `__contains__` and `__getitem__`. Internally, this must use a mapping for storage.

betty/collections.py
```diff
@@ -0,0 +1,45 @@
+"""
+Collection data tpes.
+"""
+
+from collections.abc import Callable, Collection, Iterable, Iterator
```
bartfeenstra: Rename the module to `collections`

betty/project/config.py
```diff
@@ -884,14 +890,33 @@ class _ProjectConfigurationGender(Gender):
...
         )
 
 
 @final
-class ProjectConfiguration(Configuration):
+# @todo Add the remaining elements.
+@ObjectDefinition["ProjectConfiguration"](
+    label=_("Project configuration"),
+    elements={
+        "name": ElementDefinition(SimpleDefinition(cls=str, label=_("Project name"))),
```
bartfeenstra: Don't repeat the word "project" so much in these labels

bartfeenstra: Subclass `ElementDefinition` to `AttrDefinition` and allow it to decorate property methods.

betty/project/config.py
```diff
@@ -884,14 +890,33 @@ class _ProjectConfigurationGender(Gender):
...
+            cls([get_full_sample(GenderPluginConfiguration).data]),
             label="Full",
             full=True,
         )
 
 
 @final
-class ProjectConfiguration(Configuration):
+# @todo Add the remaining elements.
+@ObjectDefinition["ProjectConfiguration"](
```
bartfeenstra: Can we make it so that we won't have to fill in the generic type here?

bartfeenstra: A default for the type var?

betty/data/aggregate/record/__init__.py
```diff
@@ -0,0 +1,127 @@
...
+
+@final
+class ElementDefinition:
+    """
+    A record element definition.
+    """
+
+    def __init__(
+        self,
+        data: DataDefinition[Any],
```
bartfeenstra: Add the element name to the definition, and in record definitions do not collect mappings but sequences of element definitions.

betty/portable/__init__.py
```diff
@@ -66,3 +69,88 @@ def dump(self) -> PortableData:
...
+
+_PortableT = TypeVar("_PortableT", bound=Portable)
+
+
+class Loader(Protocol[_DataT, _PortableDataT]):  # type: ignore[misc]
+    """
+    A callable that can load portable data.
+    """
+
+    def __call__(self, portable: _PortableDataT, /) -> _DataT:
```
bartfeenstra: Loaders should always accept `PortableData`

betty/portable/__init__.py
```diff
@@ -66,3 +69,88 @@ def dump(self) -> PortableData:
...
+        """
+
+
+class Porter(ABC, Generic[_DataT, _PortableDataT]):
+    """
+    An object capable of dumping and loading data to and from portable data.
+    """
+
+    @abstractmethod
+    def load(self, portable: _PortableDataT) -> _DataT:
```
bartfeenstra: Loaders should always accept `PortableData`

betty/data/__init__.py
```diff
@@ -1,3 +1,214 @@
...
+            yield sample()
+
+    def load(self, portable: PortableData, /) -> _DataT:
+        """
+        Create a new data instance from ``portable``.
+
+        :raises betty.exception.HumanFacingException: Raised if the portable data is invalid.
+        """
+        if self._porter:
+            return self._porter.load(portable)
```
bartfeenstra: Combine the checks in these two methods into a single helper with more specific instructions

betty/data/__init__.py
```diff
@@ -1,3 +1,214 @@
...
+    def dump(self, data: _DataT) -> PortableData:
+        """
+        Dump the data to portable data.
+        """
+        if self._porter:
+            return self._porter.dump(data)
+        raise NotPortable(f"{self} cannot be dumped because no porter was defined.")
+
+
+class HasData:
```
bartfeenstra: Make generic over the data **definition** type

bartfeenstra: But why?

betty/data/aggregate/record/object.py
```diff
@@ -0,0 +1,18 @@
...
+
+from typing import TypeVar
+
+from betty.data.aggregate.record import RecordDefinition
+from betty.data.indicator.selector import Attr
+
+_DataT = TypeVar("_DataT")
+
+
+class ObjectDefinition(RecordDefinition[_DataT, Attr]):
```
bartfeenstra, line 15: Add a `HasDataObjectDefinition` that is not actually a definition itself but that is a decorator that creates and sets an `ObjectDefinition` as soon as it is invoked as a decorator. Then make `DataDefinition`'s `cls` arg required.

betty/data/enum.py
```diff
@@ -0,0 +1,38 @@
...
+from betty.data import DataDefinition
+from betty.portable import StaticPorter
+
+if TYPE_CHECKING:
+    from betty.locale.localizable import LocalizableLike
+
+_EnumT = TypeVar("_EnumT", bound=Enum)
+
+
+class EnumDefinition(DataDefinition[_EnumT]):
```
bartfeenstra, line 20: Make final

betty/data/aggregate/record/object.py
bartfeenstra: Make both definitions final

betty/data/aggregate/record/mapping.py
```diff
@@ -0,0 +1,22 @@
...
+from collections.abc import MutableMapping
+from typing import Any, TypeVar
+
+from betty.data.aggregate.record import RecordDefinition
+from betty.data.indicator.selector import Key
+
+_MutableMappingT = TypeVar("_MutableMappingT", bound=MutableMapping[str, Any])
+
+
+class TypedMappingDefinition(RecordDefinition[_MutableMappingT, Key]):
```
bartfeenstra, line 17: Make final

betty/data/aggregate/collection/sequence.py
```diff
@@ -0,0 +1,55 @@
...
+if TYPE_CHECKING:
+    from betty.data import DataDefinition
+    from betty.locale.localizable import LocalizableLike
+
+
+_DataItemT = TypeVar("_DataItemT")
+_MutableSequenceT = TypeVar("_MutableSequenceT", bound=MutableSequence[Any])
+
+
+class SequenceDefinition(CollectionDefinition[_MutableSequenceT]):
```
bartfeenstra: Make final

betty/data/aggregate/collection/mapping.py
```diff
@@ -0,0 +1,54 @@
...
+
+if TYPE_CHECKING:
+    from betty.data import DataDefinition
+    from betty.locale.localizable import LocalizableLike
+
+_DataItemT = TypeVar("_DataItemT")
+_MutableMappingT = TypeVar("_MutableMappingT", bound=MutableMapping[str, Any])
+
+
+class MappingDefinition(CollectionDefinition[_MutableMappingT]):
```
bartfeenstra: Make final

betty/data/aggregate/__init__.py
```diff
@@ -0,0 +1,37 @@
+"""
+Aggregate data types.
```
bartfeenstra, line 2: Remove this entire file and move everything in `betty.data.aggregate` one module up.

betty/sphinx/extension/betty.py
```diff
@@ -365,11 +368,88 @@ def run(self) -> list[nodes.Node]:
...
+        data = cls.data()
+        content = ""
+
+        if isinstance(data, RecordDefinition) and data.elements:
+            content += """
+Data
+====
+
+"""
+            for element in data.elements:
```
bartfeenstra: Follow collection definitions and render those, until encountering any non-collection.

betty/data/__init__.py
```diff
@@ -1,3 +1,254 @@
...
+
+    from ty_extensions import Intersection
+
+    from betty.locale.localizable import Localizable, LocalizableLike
+    from betty.service.level import ServiceLevel
+
+_DataT = TypeVar("_DataT", default=Any)
+
+
+class Hydratable(ABC):
```
bartfeenstra: Do we definitely want to keep this here? Or is it better suited elsewhere? It's very generic and not at all specific to the dataAPI.

betty/project/config.py
```diff
@@ -880,208 +814,216 @@ class _ProjectConfigurationGender(Gender):
...
-            cls([get_full_sample(GenderPluginConfiguration).configuration]),
+            cls([get_full_sample(GenderPluginConfiguration).data]),
             label="Full",
             full=True,
         )
 
 
 @final
-class ProjectConfiguration(Configuration):
+@ObjectDefinition(
```
bartfeenstra, line 818: Add descriptions to the fields based on the old docstring

betty/data/__init__.py
bartfeenstra: It's a service API thing, innit?

betty/data/aggregate/record/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+        fields: Sequence[FieldDefinition[_ElementT, Any]],
+        description: LocalizableLike | None = None,
+        samples: Iterable[Callable[[], Sample[_DataT]]] | None = None,
+        factory: Callable[..., _DataT] | None = None,
+    ):
+        super().__init__(
+            cls=cls,
+            label=label,
+            description=description,
+            fallback_porter=StaticPorter(self._load, self._dump),
```
bartfeenstra: Set the regular ported because we only ever use the hardcoded porter created here

bartfeenstra: We could subclass porter here with one that types its portable data as mappings

betty/portable/__init__.py
```diff
@@ -60,9 +63,94 @@ def load(cls, portable: PortableData, /) -> Self:
...
+
+    @abstractmethod
+    def dump(self, data: _DataT, /) -> PortableData:
+        """
+        Dump data to its portable form.
+        """
+
+
+@final
+class StaticPorter(Porter[_DataT]):
```
bartfeenstra: `CallbackPorter`

PR #3484, opened 2026-01-09
Replace basedmypy with ty

### LINKED ISSUE #3069: Replace basedmypy with ty
basedmypy has been deprecated and is no longer keeping up with new support in mypy. There are other type checkers, but Betty specifically used basedmypy for its intersection support, so we need an alternative that supports it too.

Can we use https://pypi.org/project/libcst/ to rewrite the intersection types in our code base automatically? And can we publish that as a separate package?

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3069


ty.toml
```diff
@@ -0,0 +1,15 @@
+[rules]
+invalid-argument-type = "ignore"
```
bartfeenstra: Finetune these

betty/gramps/loader.py
```diff
@@ -510,6 +511,7 @@ async def _load_tree(self, tree: ElementTree.ElementTree) -> None:
         self._tree = tree
 
         database = self._tree.getroot()
+        assert database
```
bartfeenstra: Assert is not none (the current code causes warnings)

PR #3505, opened 2026-01-15
Extract portable data functionality into a light-weight API

bartfeenstra:


betty/portable.py
```diff
@@ -0,0 +1,73 @@
...
+PortableMapping: TypeAlias = MutableMapping[str, _PortableDataT]
+"""
+A key-value mapping of portable data.
+
+Keys are strings.
+"""
+
+
+@final
+class NotDumpable(HumanFacingException):
```
bartfeenstra: Move this to a submodule so the main API module does not depend on any other Betty functionality.

PR #3547, opened 2026-01-20
Add data definitions for plugin instance configuration

bartfeenstra:


betty/data/aggregate/record/__init__.py
```diff
@@ -167,13 +167,25 @@ def __init__(
...
+            # @todo - Make _RecordPorter public and easily reusable
+            # @todo - Then PluginInstanceConfiguration **can** provide a porter
+            # @todo
+            # @todo
+
+
+
+            # @todo Oh this default porter thing is cool but won't work for decorated objects...
+            # @todo
+            # @todo
```
bartfeenstra: Add a `porter` property we can override in subclass definitions to provide fallback behavior

betty/plugin/config/__init__.py
```diff
@@ -423,27 +434,33 @@ def dump(self) -> PortableData:
             else configuration,
         }
 
+    @override
+    def dump_key(self, key: Attr, /) -> tuple[str, PortableData]:
+        # @todo this must support defined data as well. Add test coverage for that too
```
bartfeenstra: Fix this

PR #3654, opened 2026-01-28
Refactor `ProjectConfiguration`'s plugin configuration collections to defined data

bartfeenstra:


betty/project/config.py
```diff
@@ -1006,6 +936,25 @@ class ProjectConfiguration(Data):
...
+        omit_load=True,
+        omit_dump=lambda data: not len(data),
+        default=lambda: KeyedCollection(
+            key=lambda data: data.id,
+            key_resolver=resolve_id,
+            value_resolver=resolve_plugin_configuration,
+        ),
+    )
+    """
+    Then extensions running within this project.
```
bartfeenstra: Fix typo

betty/project/config.py
```diff
@@ -965,6 +851,24 @@ class ProjectConfiguration(Data):
     The project-wide copyright notice.
     """
 
+    copyright_notices = Property(
+        KeyedCollectionDefinition(
+            item=PluginConfigurationDefinition(CopyrightNoticeDefinition),  # ty:ignore[invalid-argument-type]
```
bartfeenstra: For all of these, make the actual configuration classes subclass Data and reference that from these properties.

bartfeenstra: Also no value resolvers!

PR #3696, opened 2026-01-31
Refactor service level requirements

### LINKED ISSUE #3676: Simplify service level requirements
## To do
- **Design a solution that does not require developers to duplicate requirements.** But also, while a class may require a Project to be used, a single method (factory, hydration) may have fewer requirements. In such cases we don't want to inject the Project into such a method but we must still error because the class-level requirement has not been met.
- Apply this solution to plugin discovery as well
- Make `@require_app` and relatives usable as a decorator for any function whose first argument is a service level
- Add `require_extension(extension: ResolvableId)`
- Can these functions just raise an exception rather than return a requirement?
- Can we instead do `ServiceContainer.require()` or will that cause circular imports?
- Decorated functions do need access to the highest service level available (even if they don't need it themselves) in case they need to forward it, such as when hydrating aggregate data: an aggregate may need an `App`, but its elements may require a `Project`. The decorators would have to expand the original `services` into two arguments: one that forwards the original service level, and one that contains the required services level.
- Simplify all the factory processes, and hydration methods
- what if a function needs multiple extensions? Or other plugins active for the service level? 
- call the new service level plugin instance functionality "service plugins" or "plugin services"

### LINKED ISSUE #3701: The Wiki extension does not accept its own configuration
`Wiki` should implement `ConfigurationDependentSelfFactory`

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3676


betty/config.py
```diff
@@ -50,57 +59,40 @@ def configuration_cls(cls) -> type[_DataT]:
...
-
-class ConfigurationDependentSelfFactory(Configurable[_DataT], ABC):
-    """
-    Create factories that require configuration.
-    """
-
     @classmethod
     @abstractmethod
-    def new_for_configuration(cls, configuration: _DataT) -> ServiceLevelTarget[Self]:
+    async def new_for_services(
```
bartfeenstra: `Project`, and in the future possibly other classes, is configurable but cannot have a factory because it has additional required init arguments

bartfeenstra: Maybe split off `Configurable.configuration` into a new `HasConfiguration` subclass?

PR #3717, opened 2026-02-01
Replace `PluginTypeRepository` and `PluginRepository` with `KeyedCollection`

### LINKED ISSUE #3715: Remove plugin repositories
## To do
- Split `KeyedCollection`'s mutability off into `MutableKeyedCollection`
- Use immutable keyed collections instead of plugin repositories

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3715

## To do
- Plugin management is really a (kind of) service. Service levels MUST initialize `PluginManager` lazily. That will also prevent the current circular imports.


betty/collections.py
```diff
@@ -24,12 +24,16 @@
 
 _T = TypeVar("_T")
 _KeyT = TypeVar("_KeyT")
+_KeyCoT = TypeVar("_KeyCoT")
```
bartfeenstra, line 27: Do we need the covariant types after all?

betty/service/level/__init__.py
```diff
@@ -113,6 +112,10 @@ def plugins(self) -> PluginManager:
         """
         The plugin manager.
         """
+        from betty.service.plugin import PluginManager
```
bartfeenstra, line 115: Can we make this a service?

PR #3748, opened 2026-02-02
Refactor service containers and service levels

### LINKED ISSUE #3747: Remove bootstrap functionality from `ServiceLevel`
`ServiceLevel` is everywhere, but cannot require any (async) bootstrapping, because `betty.service.level.universe` uses it.

Currently it inherits a lot of bootstrap and shutdown code which it really shouldn't have (again, for `universe`)

Remember:
- universe: yes level, but not container
- app: yes level and container
- project: yes level and container
- extension: not level but yes container

### LINKED ISSUE #3786: `require_extension()` should keep the service level
`require_extension()` should keep the service level. That way we can stack multiple decorators:

```python
class MyClass(Manufacturable):
    @override
    @classmethod
    @require_extension(FooExtension)
    @require_extension(BarExtension)
    async def new(
        cls, project: Project, foo: FooExtension, bar: BarExtension,/
    ) -> Self:
        pass
```

`TypeVarTuple` is meant to solve this kind of issue but `Concatenate` can include either that or a `ParamSpec`, but not both.

When we refactor service containers, if we somehow make them generic over the other containers they can access (upstream service levels or extensions), we can instead refine the service level argument type using `Intersection`, so each decorator can add capabilities.

## Blocked by
- https://github.com/bartfeenstra/betty/pull/3748
- https://github.com/astral-sh/ty/issues/156

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3747

## To do
- Merge `Bootstrapped` and `Shutdownable`, and see if the combined functionality can be moved into another class if we do not use it elsewhere. Except... Let's just combine them and keep them in their own class. We can later use that to make any service level plugin (such as extensions currently) managed by their owning service level. 
- Split `ShutdownStack` into a universal and an ephemeral variant. Expose this on all service levels. 
- Fix `@todo`s
- ensure configuration is hydrated if Bootstrappable is Configurable 
- either here or in a separate PR: can we use `TypeVarTupñe` and the builder pattern to create or just either to create service containers that mesh/proxy several other containers and expose all cumulative services and their typed `@service` definitions using a single flat interface? Then we can type on things like `ServiceContainer[Intersection[Project, Maps, Raspberry mint]]`. This would help fix https://github.com/bartfeenstra/betty/issues/3786
- Can we automatically inject services? Using decorators? By requiring type hints like `jinja: Service[Environment]`? The point is that if everything gets services injected automatically, nothing needs to know about which service container something comes from 
- If we automatically inject services, we can first collect all necessary service IDs, then get the services themselves concurrently (statistically some will have asynchronous factories), and then inject them into the factory method.
- Can we use `__instancecheck__()` to pretend that a container is an instance of all the containers it is composed of?

### DIFF
--- a/betty/extension/__init__.py
+++ b/betty/extension/__init__.py
@@ -11,7 +11,7 @@ from betty.locale.localizable.gettext import _, ngettext
 from betty.plugin import Plugin, PluginTypeDefinition, ResolvableId
 from betty.plugin.dependent import DependentPluginDefinition
 from betty.plugin.discovery.entry_point import EntryPointDiscovery
-from betty.service.container import ServiceContainer
+from betty.service.container import EphemeralServiceContainer
 from betty.service.level import ServiceLevel
 from betty.typing import private
 
@@ -29,12 +29,13 @@ _ServiceLevelCoT = TypeVar(
 
 
 class Extension(
-    ServiceContainer, Plugin["ExtensionDefinition"], Generic[_ServiceLevelCoT]
+    EphemeralServiceContainer, Plugin["ExtensionDefinition"], Generic[_ServiceLevelCoT]
 ):
     """
     Integrate custom services with a :py:class:`service level <betty.service.level.ServiceLevel>`.
     """
 
+    # @todo Remove this now? Or generalize to any ServiceLevel?
     @private
     def __init__(self, *, services: _ServiceLevelCoT):
         super().__init__()
--- a/betty/service/container.py
+++ b/betty/service/container.py
@@ -4,11 +4,10 @@ Service containers.
 
 from __future__ import annotations
 
-from _warnings import warn
 from abc import abstractmethod
 from collections.abc import Awaitable, Callable
 from functools import update_wrapper
-from inspect import getmembers, iscoroutinefunction
+from inspect import iscoroutinefunction
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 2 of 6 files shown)


betty/extension/__init__.py
```diff
@@ -4,50 +4,36 @@
...
 _T = TypeVar("_T")
 
 
-class Extension(ServiceContainer, Plugin["ExtensionDefinition"]):
+class Extension(EphemeralServiceContainer, Plugin["ExtensionDefinition"]):
     """
     Integrate optional functionality with Betty :py:class:`projects <betty.project.Project>`.
     """
 
+    # @todo Remove this now? Or generalize to any ServiceLevel?
```
bartfeenstra, line 38: Expose a `services` property

betty/service/container.py
```diff
@@ -18,47 +17,48 @@
...
-
 
 _T = TypeVar("_T")
 _ServiceT = TypeVar("_ServiceT")
 _ServiceGetT = TypeVar("_ServiceGetT")
 
 
-@internal
-class ServiceContainer(Bootstrapped, Shutdownable):
+# @todo Do we need this still?
```
bartfeenstra, line 45: Check this. `@service` is generic over this though.

betty/service/container.py
```diff
@@ -186,31 +173,34 @@ def _service(
...
+class ServiceManager(Generic[_ServiceContainerT, _ServiceGetT, _ServiceT]):
     """
     Manages a single service for a service container.
     """
 
     def __init__(
         self,
         factory: Intersection[
-            ServiceFactory[_ServiceProviderT, _ServiceGetT], FunctionType
+            # @todo If we use __set_name__(), we can accept more types than just functions.
```
bartfeenstra, line 201: Follow-up?

betty/service/level.py
```diff
@@ -96,24 +82,23 @@ async def _new_target(
...
         """
-        return self._plugins
+        return ServiceLevelPluginManager(self)
+
+
+# @todo Finetune the name
+# @todo
+# @todo
+# @todo
+# @todo
```
bartfeenstra, line 121: Fix this

bartfeenstra:
Closing this in favor of https://github.com/bartfeenstra/betty/pull/3797 and https://github.com/bartfeenstra/betty/pull/3801

PR #3782, opened 2026-02-04
Decouple App from its configuration

### LINKED ISSUE #3763: Remove `App.configuration`
## To do
- Remove the property 
- Inject the individual configuration attributes as `__init__()` arguments

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3763

### DIFF
--- a/betty/app/__init__.py
+++ b/betty/app/__init__.py
@@ -21,10 +21,9 @@ from betty.cache.no_op import NoOpCache
 from betty.dirs import CACHE_DIRECTORY_PATH
 from betty.http_client import ClientErrorToUserMessageMiddleware
 from betty.http_client.rate_limit import RateLimitDefinition, RateLimitMiddleware
-from betty.importlib import fully_qualified_name
 from betty.life_cycle import LifeCycle
 from betty.life_cycle.manage import ManagedLifeCycle
-from betty.locale import DEFAULT_LOCALE
+from betty.locale import DEFAULT_LOCALE, ResolvableLocale, resolve_locale
 from betty.locale.localize import Localizer, LocalizerRepository
 from betty.locale.translation import (
     DEFAULT_TRANSLATION_REPOSITORY,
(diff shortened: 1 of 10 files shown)


betty/service/level.py
```diff
@@ -106,10 +106,8 @@ async def new_target(
             )
         if not isinstance(configuration, Data):
             configuration = target.configuration_cls().data().porter.load(configuration)
-        return await target.new(
-            services=self,
-            configuration=configuration,
-        )
+        await configuration.data().hydrate(services=self, data=configuration)
```
bartfeenstra: This new hydration call seems to make Betty freeze, going through `AggregateDefinition` somewhere...

bartfeenstra: `@require_extension("raspberry-mint")`

bartfeenstra: Let's get the decorator PR merged before we continue debugging this

betty/app/__init__.py
```diff
@@ -122,11 +103,14 @@ async def new_from_environment(cls) -> AsyncIterator[Self]:
             configuration = AppConfiguration.data().porter.load(
                 (await assert_load_file())(config.CONFIGURATION_FILE_PATH)
             )
+            await configuration.data().hydrate(services=universe, data=configuration)
```
bartfeenstra: Less relevant for App, but Project has configuration that depends on information coming from the project itself, e.g. content provider configuration that includes other content providers. Or when a user defines a custom license in their project configuration that they also want to use as the project-wide license. The license does not exist in any other service level. Chicken-and-egg all over again.

Also, for now, does App really need to hydrate its config? It's just a Locale and any validation issues are resolved when loading portable data.

bartfeenstra: I guess if either App or Project needs any kind of hydration for parts of their configuration, they should do so when bootstrapping?

bartfeenstra: However, hydrating AppConfiguration or ProjectConfiguration in its entirety allows us to raise errors with useful indicators as to where the problem originates. If we hydrate only nested parts, we do not have that information at all.

TLDR;configuration validation must happen on the entirety of the data.

Now, this is a more generic problem. Theoretically extensions can have the same issue.

betty/app/__init__.py
```diff
@@ -86,32 +81,18 @@ def __init__(
 
         cls = type(self)
         super().__init__()
-        self._configuration = configuration
+        self._locale = None if locale is None else resolve_locale(locale)
```
bartfeenstra: DEFAULT_LOCALE. And remove the default further down

betty/app/data.py
```diff
@@ -39,21 +38,10 @@ class AppConfiguration(Data):
...
-    def locale(self) -> Locale | None:
-        """
-        The application locale.
-        """
-        return self._locale
-
-    @locale.setter
-    def locale(self, locale: Locale | None) -> None:
-        self._locale = locale
+    locale = Optional(Property(LocaleDefinition()))
```
bartfeenstra: Add `LocaleProperty`. It must set a resolver. Use it for project locale configuration as well. Check for other usages of `LocaleDefinition`.

PR #3821, opened 2026-02-10
Refactor `PluginConfiguration` into `PluginManufacturer`

### LINKED ISSUE #3817: Rename `PluginConfiguration` to `PluginManufacturer`
It **is** specific to plugins, but we don't call things configuration anymore. Also this class can be used to create new instances even if they do not use data ('configuration'), so find a suitable name.

Implement `__call__()` using a factory compatible signature. 

We may have to subclass this per plugin type, but really that should be fine, I guess?

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3817


betty/plugin/config/__init__.py
```diff
@@ -137,46 +137,56 @@ def __init__(
...
 ):
     """
     Configure a single plugin instance.
 
-    Use this with :py:class:`betty.plugin.data.PluginConfigurationDefinition` to provide defined data.
+    Use this with :py:class:`betty.plugin.data.PluginManufacturerDefinition` to provide defined data.
     """
 
+    # @todo
+    PLUGIN_TYPE: type[_PluginDefinitionT]
```
bartfeenstra: Make this an abstract class method now we access this from outside the class as well (`PluginManufacturerDefinition`)

betty/plugin/config/__init__.py
```diff
@@ -190,43 +200,49 @@ def load(cls, portable: PortableData, /) -> Self:
...
-        return self.id, {} if self.configuration is Void() else {
-            "configuration": self._dump_configuration(self.configuration),  # ty:ignore[invalid-argument-type]
+        return self.id, {} if self.data is Void() else {
+            "configuration": self._dump_configuration(self.data),  # ty:ignore[invalid-argument-type]
         }
 
-    async def new_plugin(
-        self, services: ServiceLevel, plugin_type: type[_PluginDefinitionT], /
-    ) -> _PluginT:
+    @final
```
bartfeenstra: Do we want to introduce a `@manufacturer` decorator that does nothing and returns the original method, but whose sole purpose is type checking? 

In a separate PR, do we want decorators for similar callbacks, such as porters, loaders, and discoveries?

If yes, put these decorators in `typing` submodules.

betty/plugin/config/__init__.py
```diff
@@ -190,43 +200,49 @@ def load(cls, portable: PortableData, /) -> Self:
...
         Create a new instance of the configured plugin.
         """
         return await services.factory.new(
-            (await services.plugins.plugins(plugin_type))[self.id].cls,
-            self.configuration,
+            (await services.plugins.plugins(self.PLUGIN_TYPE))[self.id].cls,
+            self.data,
         )
 
+    # @todo Do we want subclasses to provide their own samples?
```
bartfeenstra: Fix this. Unique samples per plugin type would not be a requirement but definitely would improve UX. Can we add a `samples` argument to `PluginManufacturerDefinition`? No! Just allow subclasses to optionally implement `Samplable`.

betty/plugin/config/__init__.py
```diff
@@ -137,46 +136,61 @@ def __init__(
...
-        id: ResolvableId[_PluginDefinitionT],  # noqa: A002
-        configuration: Data | PortableData | Void = Void(),  # noqa: B008
+        plugin: ResolvableId[_PluginDefinitionT],
+        data: Data | PortableData | Void = Void(),  # noqa: B008
         /,
     ):
         super().__init__()
-        self._id = assert_machine_name()(resolve_id(id))
-        self._configuration = configuration
+        self._plugin = assert_machine_name()(resolve_id(plugin))
```
bartfeenstra: Do not resolve the ID. Do validate the ID if it's a string.

betty/plugin/config/__init__.py
```diff
@@ -137,46 +136,61 @@ def __init__(
...
     @override
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, type(self)):
             return NotImplemented
-        return (self.id, self.configuration) == (other.id, other.configuration)
+        return (self.id, self.data) == (other.id, other.data)
 
+    @final
     @property
     def id(self) -> MachineName:
```
bartfeenstra: The return value is `ResolvableId`

betty/plugin/config/__init__.py
```diff
@@ -190,43 +204,96 @@ def load(cls, portable: PortableData, /) -> Self:
...
             "configuration": self._dump_configuration(configuration),
         }
 
+    @final
     @override
     def dump_key(self, key: Attr, /) -> tuple[str, PortableData]:
-        return self.id, {} if self.configuration is Void() else {
-            "configuration": self._dump_configuration(self.configuration),  # ty:ignore[invalid-argument-type]
+        return self.id, {} if self.data is Void() else {
+            "configuration": self._dump_configuration(self.data),  # ty:ignore[invalid-argument-type]
```
bartfeenstra: Do we want to rename this field to `data`? It would be more consistent from an API perspective but perhaps less intuitive for users who just edit their project configuration files.

betty/service/factory.py
```diff
@@ -42,6 +41,9 @@ async def new(cls, services: ServiceLevel, /) -> Self:
 _ManufacturableT = TypeVar("_ManufacturableT", bound=Manufacturable)
 
 
+Manufacturer: TypeAlias = Callable[[ServiceLevel], Awaitable[_T] | _T]
```
bartfeenstra: We need to allow callables without arguments as well, because that is what classes are.

betty/tests/service/test_factory.py
```diff
@@ -11,11 +11,11 @@ class _TargetType:
     pass
 
 
-def _sync_callable_target() -> _TargetType:
+def _sync_callable_target(services: ServiceLevel, /) -> _TargetType:
```
bartfeenstra: Add targets without arguments as well.

betty/service/factory.py
bartfeenstra: Classes with initializers without arguments, that is.

betty/plugin/config/__init__.py
bartfeenstra: Nah

betty/service/factory.py
bartfeenstra: Without **required** arguments

betty/plugin/config/__init__.py
```diff
@@ -126,56 +120,89 @@ def __init__(
...
-_PluginConfigurationPluginDefinitionT = TypeVar(
-    "_PluginConfigurationPluginDefinitionT", bound=PluginDefinition
+_PluginManufacturerPluginDefinitionT = TypeVar(
+    "_PluginManufacturerPluginDefinitionT", bound=PluginDefinition
 )
 
 
-@final
-class PluginConfiguration(
+class PluginManufacturer[
```
bartfeenstra: Implement `Data.data()`. Decorate it with `@lru_cache`. Lazily create the definition using `cls.type()`. Remove the definition class.

betty/plugin/config/__init__.py
```diff
@@ -126,56 +120,89 @@ def __init__(
...
+        data: Data | PortableData | Void = Void(),  # noqa: B008
         /,
     ):
         super().__init__()
-        self._id = resolve_id(id)
-        self._configuration = configuration
+        self._plugin_id = resolve_id(plugin)
+        self._plugin_data = data
+
+    def __hash__(self):
```
bartfeenstra: Make final

bartfeenstra: Add test coverage for the three different kinds of plugin data.

PR #3834, opened 2026-02-16
Make `MachineName` a class

### LINKED ISSUE #3832: Make `MachineName` a class
And subclass `str`. Merge `assert_machine_name()` into `__new__()`

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3832

### DIFF
--- a/betty/machine_name.py
+++ b/betty/machine_name.py
@@ -1,53 +1,115 @@
 """
-Define portable machine names.
+Machine names.
 """
 
 from __future__ import annotations
 
 import re
-from typing import Any, TypeAlias, TypeGuard, final
+from typing import TYPE_CHECKING, Self, TypeAlias, final
 
-from betty.assertion import AssertionChain, assert_str
-from betty.data import DataDefinition
+from typing_extensions import override
+
+from betty.assertion import assert_str
+from betty.data import Data, DataDefinition
+from betty.data.aggregate.record.object.property import Property
 from betty.exception import HumanFacingException
-from betty.functools import passthrough
 from betty.locale.localizable.gettext import _
-from betty.portable import CallbackPorter
+from betty.locale.localizable.markup import Paragraph
+from betty.portable import Portable, PortableData
 
-MachineName: TypeAlias = str
-"""
-A machine name is a string that meets these criteria:
-- At most 250 characters long.
-- Lowercase letters, numbers, and hyphens (-).
+if TYPE_CHECKING:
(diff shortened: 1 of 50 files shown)


betty/machine_name.py
```diff
@@ -1,90 +1,125 @@
...
+    @override
+    def __new__(cls, machine_name: str, /):
+        if _MACHINE_NAME_PATTERN.fullmatch(machine_name) is None:
+            raise InvalidMachineName(machine_name)
+        return super().__new__(cls, machine_name)
+
+    def __init__(self, machine_name: str, /):
+        pass
+
+    def __hash__(self):
```
bartfeenstra: Ensure this returns exactly the same has as hashing a `str` with the same value would produce. We can then remove extraneous `str(...)` calls we added in this PR when using machine names as mapping keys.

betty/machine_name.py
```diff
@@ -1,90 +1,125 @@
...
+        return hash((type(self), str(self)))
+
+    @override
+    @classmethod
+    def load(cls, portable: PortableData, /) -> Self:
```
bartfeenstra, line 55: Remove this. Instead, update DataDefinition to create a pass through porter if the target data type is portable data already. Do this in a separate PR. Remove all porters from definitions of data that is portable already.

betty/machine_name.py
```diff
@@ -1,90 +1,125 @@
...
-A machine name is a string that meets these criteria:
-- At most 250 characters long.
-- Lowercase letters, numbers, and hyphens (-).
+if TYPE_CHECKING:
+    from betty.locale.localizable import ResolvableLocalizable
 
-See :py:func:`betty.machine_name.validate_machine_name`.
-"""
+_MACHINE_NAME_DESCRIPTION = _(
+    "An identifier of at most 250 characters long, made up of lowercase letters, numbers, and/or hyphens (-)."
```
bartfeenstra: "A machine name is an identifier..."

betty/machine_name.py
```diff
@@ -1,90 +1,125 @@
...
     """
     Raised when something is not a valid machine name.
     """
 
     def __init__(self, value: str, /):
         super().__init__(
-            _(
-                '"{value}" is not a valid machine name: only lowercase letters, digits, and hyphens (-) are allowed.'
-            ).format(value=value)
+            _('"{value}" is not a valid machine name: {details}.').format(
```
bartfeenstra: No colon. No formatting. Use Paragraph to join the two localizables.

PR #3884, opened 2026-02-23
Refactor plugin management

### LINKED ISSUE #3715: Remove plugin repositories

### LINKED ISSUE #3772: Make `PluginManager.plugins` asynchronous
Its predecessor used to be asynchronous, but at the time we had too much synchronous calling code, so we made it synchronous. Check the current state of things and if we can make it asynchronous again, in which case use a lock and lazily discover plugins.

Also allow discoveries to only discover plugin IDs, and lazily load each plugin's definition upon request. This improves perform for things like Spdx licenses, which come in bulk but of which the vast majority is never used.

bartfeenstra:


betty/itertools.py
```diff
@@ -0,0 +1,12 @@
+"""
+Tools for iterables.
+"""
+
+from collections.abc import AsyncIterable, Iterator
+
+
+async def aiter_to_iter[T](values: AsyncIterable[T]) -> Iterator[T]:
```
bartfeenstra: Remove this and replace calling code with `[x async for x in ...]`

PR #3889, opened 2026-02-27
Remove `PluginTypeDefinition.discovery`

### LINKED ISSUE #3714: Consider automatically using entry point discovery for all plugin types
Unless there is a reason why some individual plugin types MUST NOT support entry point discovery?

bartfeenstra:


betty/console/command/__init__.py
```diff
@@ -41,10 +39,6 @@ async def configure(self, parser: argparse.ArgumentParser) -> CommandFunction:
     label=_("Command"),
     label_plural=_("Commands"),
     label_countable=ngettext("{count} command", "{count} commands"),
-    discovery=[
-        EntryPointDiscovery("betty.command"),
-        *([EntryPointDiscovery("betty.dev.command")] if about.IS_DEVELOPMENT else []),
```
bartfeenstra, line 46: Replace the dev command discovery with an entry point that only discovers definitions if Betty is installed in development mode

bartfeenstra: Can we add a `@require_dev` decorator?

PR #3894, opened 2026-03-01
Improve the plugin API's resolvers

bartfeenstra:


betty/plugin/__init__.py
```diff
@@ -149,23 +150,51 @@ def plugin(cls) -> PluginDefinitionT:
 """
 
 
-def resolve_definition[PluginDefinitionT: PluginDefinition](
-    definition: ResolvableDefinition[PluginDefinitionT], /
+@overload
+def resolve_plugin_definition[PluginDefinitionT: PluginDefinition](
```
bartfeenstra, line 186: Don't chain resolvers. Instead just raise `ValueError`.

PR #3914, opened 2026-03-14
Simplify the ordered plugin API

### LINKED ISSUE #3911: Shorten `comes_before` and `comes_after` to `before` and `after`

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3911


betty/deriver.py
```diff
@@ -133,12 +133,12 @@ async def _derive_person(
             else:
                 return 0, 0
 
-        # Aggregate event type order from references and backreferences.
```
bartfeenstra, line 136: We no longer include backreferences....

betty/service/plugin.py
```diff
@@ -286,24 +287,34 @@ async def _new_plugin_collection[ServicePluginDefinitionT: ServicePluginDefiniti
...
-            plugins_by_id = {plugin.plugin().id: plugin for plugin in plugins}
-            sorter = await sort_ordered_plugin_graph(
-                self._services.plugins[plugin_type],
-                (plugin.plugin() for plugin in plugins),
-            )
+            sorter = TopologicalSorter[MachineName]()
+            for plugin in plugins:
+                plugin_definition = plugin.plugin()
+                assert isinstance(plugin_definition, OrderedPluginDefinition)
+                sorter.add(plugin_definition.id)
```
bartfeenstra, line 296: Move the creation of a sorter back to an API function because we will need it in Deriver as well.

betty/deriver.py
```diff
@@ -68,12 +69,17 @@ async def derive(self) -> None:
         """
         Derive additional data.
         """
+        sorter = new_sorter(
```
bartfeenstra: Combine these two into a single line as we don't need the sorter for anything else.

betty/deriver.py
```diff
@@ -68,12 +70,49 @@ async def derive(self) -> None:
...
+                # @todo Let's do this for BEFORE for now, then expand
+                # @todo Also, remember that order is bidrectional. We can only discard  an event type
+                # @todo where before() is None if there is no direct inverse chain from the origin to it.
+                # @todo
+                # @todo We may need to prepare the graph as a mapping first, and once all edges have been added, we
+                # @todo remove unconnected nodes?
+                # @todo How?
+                # @todo - Two graphs, one to model all BEFORE one to model all AFTER relationships
+                # @todo - Follow both graphs to see if the orgin is connected to the given event type.
+                # @todo - If no connection is found in either graph, do not include the given event type.
```
bartfeenstra: Actually we need a single graph, because BEFORE and AFTER relationships combined can be a valid link

bartfeenstra: Also BEFORE is inverted AFTER and vice versa

betty/service/plugin.py
bartfeenstra: Actually we did not end up using the API method more than once, so revert this change.

PR #3946, opened 2026-03-16
Replace `CssProvider` with CSS resource plugins

### LINKED ISSUE #3671: HTML plugins
## To do
- Merge `CssProvider` and `JsProvider` into `Html` (is that a good name?)
- Consider refactoring each CSS and JS asset into individual plugins. Make them ordered.

## Blocked by
- https://github.com/bartfeenstra/betty/issues/3670

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3671

### DIFF
--- a/assets/universe/locale/ar/betty.po
+++ b/assets/universe/locale/ar/betty.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Betty VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2026-03-16 19:27+0000\n"
+"POT-Creation-Date: 2026-03-17 00:17+0000\n"
 "PO-Revision-Date: 2026-02-17 18:12+0000\n"
 "Last-Translator: Bart Feenstra <bart@bartfeenstra.com>\n"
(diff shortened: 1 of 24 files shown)


betty/plugins/css_resource/webpack_entry_point.py
```diff
@@ -0,0 +1,40 @@
...
+from betty.extension import Extension
+from betty.html.css import CssResource, CssResourceDefinition
+from betty.plugin.discovery import ResolvableDiscovery
+from betty.plugins.extension.webpack import EntryPointProvider
+from betty.project import Project
+from betty.service.requirement.project import require_project
+
+
+@require_project
+async def webpack_entry_point(
```
bartfeenstra: We create one plugin per entry point, but in practice we use them all together anyway. Also, currently, we cannot require these discovered plugins anywhere.

bartfeenstra: Also I believe this is a leftover from early attempts to be able to lazily load CSS files per entry point. I don't think we ended up doing that at all (verify!). If we don't, we can use a single static resource here (undo the changes that allowed multiple resources) once we have refactored the We pack builds to aggregate all CSS into a single file.

PR #3978, opened 2026-03-22
Reintroduce a requirement API

### LINKED ISSUE #3912: Validate service plugin requirements
Service plugins can define requirements on plugin types that are not supported as service plugins, either at all or in the given service provider. Raise an error when this happens. Add test coverage. 

Also allow plugins that are not themselves service plugins to declare requirements on service plugins. We want to be able to ensure that things like content providers in Raspberry Mint's regions can declare their requirements and that those are checked, preferably when bootstrapping the relevant service level (in the case of extensions) so we can attempt to enable required service plugins automatically. 

Currently requirements are only checked against the given service level. However, service levels can now be chained, so if the given service level does not provide the required service plugin type, check the upstream.

Can we allow requirements to be declared on service levels as well? E.g. when a plugin needs an app, project, or extension?

### LINKED ISSUE #3952: Merge `require_app()` and `require_project()`
Once we have untangled our plugins and we won't have to worry about circular imports anymore, merge these two functions.

Additionally, allow them to accept an argument that is a mixin class. Finally, allow extensions to be specified. Should they have an upstream too? Should we refactor `ChainedServiceLevel` into `ChainedServiceProvider`?

## Blocked by
- https://github.com/bartfeenstra/betty/issues/3458

### LINKED ISSUE #3974: Simplify requirements
Requirements are currently a mapping that has limited typing abilities. It also does not scale to other types of requirements. 

## To do
- Create a `Requirement` base class with `async Requirement.check(self, services: ServiceLevel, /)`
- When declaring requirements, allow manufacturables as well. That way requirements can optionally access services.
- How to chain requirements, or express requirements' own requirements? E.g. if one requirement is for a project, it may depend on another requirement for an app? Or one requirement that needs a service may depend on another requirement for the service's service provider.
- Make `ServicePluginDefinition.requires` `Iterable[ResolvableServiceRequirement]` where `type ResolvableServiceRequirement = Requirement | type[ServiceLevel] | ServicePluginDefinition | type[Plugin[ServicePluginDefinition]] | tuple[type[ServicePluginDefinition], MachineName]`. Or can we make this fully typed and not support plugin IDs/`MachineName` here at all? We shouldn't have cyclic dependencies once we're done refactoring all our plugins.
- Add concrete subclasses to easily add `App`, `Project`, or any service plugin by its class.
- Add a concrete final subclass that checks if Node.js and npm are available and that caches its result.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3974, https://github.com/bartfeenstra/betty/issues/3952, and https://github.com/bartfeenstra/betty/issues/3912


betty/requirement/__init__.py
```diff
@@ -16,17 +17,28 @@
...
+    """
+    Decorate a callable to check a requirement.
+    """
 
-        self._require = require
+    def __init__(self, requirement: Requirement[RequirementT], /):
+        self._requirement = requirement
 
     @overload
     def __call__(self, f: ServiceLevel, target: Any, /) -> Awaitable[RequirementT]:
```
bartfeenstra: Remove this overload because instances of this class are no longer meant to be called directly with a service level.

betty/requirement/__init__.py
```diff
@@ -48,29 +60,25 @@ def __call__[**P, ReturnT, MagicT](
 
     def __call__[**P, ReturnT](
         self, f, target=None
-    ) -> CallableRequirement[RequirementT, P, ReturnT] | Awaitable[RequirementT]:
+    ) -> _CallableRequire[RequirementT, P, ReturnT] | Awaitable[RequirementT]:
         """
         Decorate a callable with this requirement.
         """
         if isinstance(f, ServiceLevel):
```
bartfeenstra: Remove support for calling this with a service level.

betty/requirement/app.py
```diff
@@ -0,0 +1,37 @@
...
+from betty.locale.localizable.gettext import _
+from betty.project import Project
+from betty.requirement import Requirement, UnmetRequirement, require
+
+if TYPE_CHECKING:
+    from betty.service.level import ServiceLevel
+
+
+@final
+class AppRequirement(Requirement[App], Singleton):
```
bartfeenstra: With `ChainedServiceLevel` we no longer need separate support for apps and projects. We can just use `type[ServiceLevel]` and resolve it.

bartfeenstra: Similarly we can use `ChainedServiceLevel` and `ServicePluginProvider.service_plugins` to check requirements for any service plugin of any type on any service level.

PR #3980, opened 2026-03-23
Check requirements for non-service plugins

### LINKED ISSUE #3979: Check requirements for non-service plugins
In order to check requirements for non-service plugins (e.g. content) at the same time as checking requirements for service plugins, we must be able to extract these requirements from `Data` (with `ProjectConfiguration` being an example root).

## To do
- Move `ServicePluginDefinition.requires` to `PluginDefinition.requires`
- Add `requires` to `ServicePluginProvider.__init__()`. This should replace `Project.__init__()`'s `extensions`
- Add an API for `Data` classes to expose requirements and a reusable implementation to to this for configured plugins.

## Blocked by 
- https://github.com/bartfeenstra/betty/pull/3978

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3979

## Thoughts
- Is this even doable at all? All use cases use `PluginManufacturer` which is inherently unable to produce a plugin's `Data` object without a service level (`ServiceLevel.plugins` is needed to get the plugin class, which exposes the data class, which we can then initialize using the portable data. Only at this point can we check if that data is `HasRequirements`.
- Even if we cannot automatically enable requirements for non-service plugins, declaring them does allow us auto enabling in tests and to automatically provide detailed error reporting.
- Consider extracting individual improvements from this PR into separate PRs and closing this one.

## To do
- Add an API for `Data` classes to expose requirements and a reusable implementation to do this for configured plugins.

### DIFF
--- a/betty/app/__init__.py
+++ b/betty/app/__init__.py
@@ -31,7 +31,11 @@ from betty.portable.file import assert_load_file
 from betty.service.factory import DataManufacturable
 from betty.service.level import ChainedServiceLevel, Plugins, ServiceLevel
 from betty.service.level.universe import UNIVERSE
-from betty.service.plugin import ServicePluginManager, ServicePluginProvider
+from betty.service.plugin import (
+    ServicePluginManager,
+    ServicePluginProvider,
+    ServicePlugins,
+)
 from betty.service.provider import ServiceFactory, service
 from betty.typing import threadsafe
 from betty.user.no_op import NoOpUser
@@ -47,11 +51,15 @@ if TYPE_CHECKING:
     from betty.plugin.discovery import ResolvableDiscovery
     from betty.user import User
 
+type AppServicePlugin = AssetDefinition | RateLimitDefinition
+
 
 @final
 @threadsafe
 class App(
-    DataManufacturable[AppConfiguration], ChainedServiceLevel, ServicePluginProvider
+    DataManufacturable[AppConfiguration],
+    ChainedServiceLevel,
+    ServicePluginProvider[AppServicePlugin],
 ):
     """
     The Betty application.
@@ -72,6 +80,7 @@ class App(
         locale: ResolvableLocale | None = None,
         plugins: Plugins | None = None,
         process_pool: futures.ProcessPoolExecutor | None = None,
+        service_plugins: ServicePlugins[AppServicePlugin] = (),
         translations: TranslationRepository | None = None,
         user: User | None = None,
     ):
@@ -99,6 +108,7 @@ class App(
             if cache_factory is None
             else cache_factory,
         )
+        self._service_plugins = service_plugins
 
     async def _bootstrap_localizer(self) -> None:
         self._user.localizer = await self.localizer
@@ -166,12 +176,13 @@ class App(
 
     @override
     @service
-    async def service_plugins(self) -> ServicePluginManager:
+    async def service_plugins(self) -> ServicePluginManager[AppServicePlugin]:
         service_plugins = ServicePluginManager(
-            {
-                AssetDefinition: (),
-                RateLimitDefinition: (),
+            plugin_types={
+                AssetDefinition,
+                RateLimitDefinition,
             },
+            plugin_manufacturers=self._service_plugins,
             services=self,
         )
         await service_plugins.bootstrap()
--- a/betty/event_type/__init__.py
+++ b/betty/event_type/__init__.py
@@ -19,8 +19,6 @@ from betty.plugin.factory import PluginManufacturer
 from betty.plugin.ordered import Order, OrderedPluginDefinition
 
 if TYPE_CHECKING:
-    import builtins
-
     from betty.locale.localizable import CountableLocalizable, ResolvableLocalizable
     from betty.machine_name import MachineName, ResolvableMachineName
     from betty.plugins.entity.person import Person
@@ -99,5 +97,5 @@ class EventTypeManufacturer(PluginManufacturer[EventTypeDefinition, EventType]):
 
     @override
     @classmethod
-    def type(cls) -> builtins.type[EventTypeDefinition]:
+    def plugin_type(cls) -> type[EventTypeDefinition]:
         return EventTypeDefinition
--- a/betty/plugin/factory.py
+++ b/betty/plugin/factory.py
@@ -29,8 +29,6 @@ from betty.sample import Samplable, Sample, Samples, Size
 from betty.typing import Void, VoidType
 
 if TYPE_CHECKING:
-    import builtins
-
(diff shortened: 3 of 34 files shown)


betty/event_type/__init__.py
```diff
@@ -99,5 +99,5 @@ class EventTypeManufacturer(PluginManufacturer[EventTypeDefinition, EventType]):
 
     @override
     @classmethod
-    def type(cls) -> builtins.type[EventTypeDefinition]:
+    def plugin_type(cls) -> builtins.type[EventTypeDefinition]:
```
bartfeenstra: Remove the `builtins` imports in all these files.

betty/app/__init__.py
```diff
@@ -72,6 +76,7 @@ def __init__(
         locale: ResolvableLocale | None = None,
         plugins: Plugins | None = None,
         process_pool: futures.ProcessPoolExecutor | None = None,
+        service_plugins: ServicePlugins | None = None,
```
bartfeenstra: Here and for Project, type this so only service plugins of supported types can be enabled.

betty/project/__init__.py
```diff
@@ -97,6 +110,7 @@ def __init__(
         self._configuration = configuration
         self._directory = directory
         self._ancestry = Ancestry() if ancestry is None else ancestry
+        self._service_plugins = service_plugins
```
bartfeenstra: Also add all service plugins from `ProjectConfiguration`.

betty/app/__init__.py
```diff
@@ -169,9 +176,10 @@ async def new_isolated(
     async def service_plugins(self) -> ServicePluginManager:
```
bartfeenstra: Extract this into a mixin base class. Use it for App and Project. Later we can use it for extensions as well.

bartfeenstra: Also make this final in the new class.

betty/service/plugin.py
```diff
@@ -213,60 +207,68 @@ def auto(self) -> bool:
...
-            else {
-                plugin_type: {plugin.plugin_id: plugin for plugin in plugins}
-                for plugin_type, plugins in service_plugin_manufacturers.items()
-            }
-        )
+        self._service_plugin_types = plugin_types
+        self._service_plugin_manufacturers = defaultdict(dict)
+        if plugin_manufacturers is not None:
+            for service_plugin_manufacturer in plugin_manufacturers:
+                if isinstance(service_plugin_manufacturer, PluginManufacturer):
```
bartfeenstra: What if we get multiple manufacturers for the same plugin?

There MUST be at most one `PluginManufacturer` per plugin. There may be zero or more 'class-based manufacturers'. `PluginManufacturer` MUST override any class. Classes MUST NOT override `PluginManufacturer`.

betty/project/data.py
```diff
@@ -550,6 +554,13 @@ def _default_license(cls) -> LicenseManufacturer:
 
         return LicenseManufacturer(AllRightsReserved)
 
+    @override
+    @property
+    def service_plugin_requirements(
+        self,
+    ) -> HasServicePluginRequirementsRequirements[ProjectServicePlugin]:
+        return self.extensions
```
bartfeenstra: Return copyright notice, license, etc. Also include the requirements for the data/configuration of each of these plugins.

betty/service/plugin.py
```diff
@@ -368,14 +374,54 @@ def __iter__(self) -> Iterator[type[ServicePluginDefinition]]:
...
+    PluginManufacturer[ServicePluginT, Plugin[ServicePluginT]]
+    | tuple[type[ServicePluginT], MachineName]
+)
+
+type HasServicePluginRequirementsRequirements[
+    ServicePluginT: ServicePluginDefinition
+] = Iterable[HasServicePluginRequirementsRequirement[ServicePluginT]]
+
+
+class HasServicePluginRequirements[ServicePluginT: ServicePluginDefinition](ABC):
```
bartfeenstra: Implement this for all content plugins that take configuration for nested content plugins.

betty/project/__init__.py
```diff
@@ -113,7 +129,32 @@ async def new(
         """
         Create a new instance.
         """
-        return cls(directory, app=app, configuration=data)
+        # @todo How do we get requirements for non-service plugins here, if we cannot access self.plugins?
```
bartfeenstra: We would need to add another init kwarg to specify the (non-service) plugins to check requirements for.

bartfeenstra: Ideally this kwarg can be used to test any requirement. Can we do something similar to `ServicePluginRequirement` but for plugins that are not themselves service plugins? `PluginRequirementsRequirement`?

betty/project/__init__.py
```diff
@@ -113,7 +129,32 @@ async def new(
...
+        return cls(
+            directory,
+            app=app,
+            configuration=data,
+            service_plugins=await gather(
+                *map(
+                    self._ensure_service_plugin,
+                    data.service_plugin_requirements,
```
bartfeenstra: Allow this to expose any requirement. Rename the base class to `HasRequirements`. Use it for `PluginDefinition`.

betty/app/__init__.py
```diff
@@ -72,6 +80,7 @@ def __init__(
         locale: ResolvableLocale | None = None,
         plugins: Plugins | None = None,
         process_pool: futures.ProcessPoolExecutor | None = None,
+        service_plugins: ServicePlugins[AppServicePlugin] = (),
```
bartfeenstra, line 83: Replace this with separate kwargs for assets and rate limits

betty/plugin/factory.py
```diff
@@ -77,9 +83,28 @@ def __hash__(self):
...
+        # @todo
+        # @todo
+        raise NotImplementedError
+        if isinstance(self.plugin_data, HasRequirements):
+            yield from self.plugin_data.requires
+
     @classmethod
     @abstractmethod
-    def type(cls) -> builtins.type[_PluginManufacturerPluginDefinitionT]:
+    def plugin_type(cls) -> type[_PluginManufacturerPluginDefinitionT]:
```
bartfeenstra, line 107: Extract the rename into a separate PR.

betty/plugins/asset/project.py
```diff
@@ -24,6 +24,7 @@ def _discover(project: Project, /) -> Iterable[ResolvableDiscovery[AssetDefiniti
         assets=project.assets_directory,
         after=lambda other: other != Universe.plugin().id,
         before={Universe},
+        auto=True,
```
bartfeenstra, line 27: Extract the project assets changes into a separate PR.

PR #4020, opened 2026-03-30
Refactor entity resolvers

### LINKED ISSUE #2778: To-many entity resolvers do not persist order
- refactor resolver so they're always added unchecked. This will allow us to keep associations ordered
- no more many resolvers? They'd be more memory efficient though (pickling)
- for to-many associations, only accept EITHER entities OR a singular resolver 
- when resolving resolvers, to not automatically update the associate's field
- for bidirectional associations, require resolvers for associates on either end

### LINKED ISSUE #3533: Merge entity references and entity resolvers
They both do somewhat similar things and with the (hopefully soon) introduction of data definitions and a generic hydration API, we may be able to merge both references and resolvers into one.

However, the core difference between the two is that references reference an existing entity in the ancestry by type and ID, and that resolvers (currently) are allowed to produce the entities themselves in whatever way they choose. One real-world reason to keep some support for this is that when loading entities, at the time associates must be filled in, it is not always known what those associates' entity IDs are.

## Blocked by
- https://github.com/bartfeenstra/betty/pull/3481
- https://github.com/bartfeenstra/betty/issues/3587

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3533

## To do
- Can we finish creating a unified solution?
- Can we fix https://github.com/bartfeenstra/betty/issues/2778 while we're at it?

### DIFF
--- a/betty/entity/association.py
+++ b/betty/entity/association.py
@@ -120,7 +120,9 @@ class _Association[OwnerT: Entity, AssociateT: Entity](LinkedDataDumper[OwnerT])
         )
 
     @abstractmethod
-    def resolve(self, owner: OwnerT, /) -> None:
+    def resolve(
+        self, entities: MultipleTypesEntityCollection, owner: OwnerT, /
+    ) -> None:
         """
         Resolve any associates the owner may have for this association.
         """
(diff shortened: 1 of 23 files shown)


bartfeenstra:
Closing in favor of https://github.com/bartfeenstra/betty/pull/4265

PR #4023, opened 2026-03-30
Make plugin services declarative

### LINKED ISSUE #3936: Service plugin cardinality
Service plugins are now a collection of zero or more plugins. For the project license and copyright notice, there must always be exactly one plugin. 

## To do
- Add a way for service providers to declare the cardinality of their service plugins:
  - one and one only, e.g. for project copyright notice and license 
  - zero or more unique plugins, keyed by ID (also batched), e.g. for extensions 
  - zero or more plugins, unrestricted, as a sequence (also batched), e.g. for content providers 
- Instead of registering service plugins using their plugin type, can we register them with a name or something that describes the purpose? That would prevent ambiguous situations where a plugin type is used for different services. We'd have to rethink how we express dependencies then.
- Consider moving `ServicePluginDefinition.requires` to `PluginDefinition.requires` so we can check requirements for plugins that are not themselves usable as services.
- Make the project license and copyright notice services.

### LINKED ISSUE #3958: Lazily initialize service plugins
Service plugins are now initialized in one action. This means that if one service plugin needs another, everything hangs.

The solution is to initialize each and every service plugin only when that plugin is needed, and to make `ServicePluginManager` re-entrant.

This is different from things like content providers in that those are not individually available services, with at most one instance per plugin.

### LINKED ISSUE #4021: Classless service plugins
Service plugins must currently always be classed. This prevent us from making certain service plugin types, which do not have instance functionality, classless, such as assets, links, and rate limits. 

For these plugin types, the services must consist of plugin definitions rather than plugin instances. 

As the service plugin API grows, we may want to refactor service plugin services to be descriptors. We can then use different types for different behaviors (instances/definitions, keyed/unkeyed, cardinality). During bootstrap, the owning class can gather all descriptors and act on those, instead of getting the relevant service configuration through initializer kwargs.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/4021, https://github.com/bartfeenstra/betty/issues/3936, and https://github.com/bartfeenstra/betty/issues/3958.

## Follow-ups
- Allow plugin service requirements to declare requirements on plugin services provided by extensions. When checking, when we have a service level, how do we get the extension? The requirement would have to declare the extension service on a service level.
- https://github.com/bartfeenstra/betty/issues/4071


betty/service/plugin/service.py
```diff
@@ -87,43 +114,104 @@ def keys(self) -> Sequence[MachineName]:
...
+
+    def __get__(self, instance, owner):
+        if instance is None:
+            return self
+        return getattr(instance, self._attr_name)
+
+
+# @todo Refactor this once we've got everything working
+@final
+class KeyedPluginService[
```
bartfeenstra: `UniquePluginService`

betty/service/plugin/service.py
```diff
@@ -87,43 +114,104 @@ def keys(self) -> Sequence[MachineName]:
...
-    ServicePluginT: ServicePluginDefinition = ServicePluginDefinition
-] = _Plugins[ServicePluginT]
-type SupportPlugins = _Plugins
+        self._type = plugin_type
+        self._init_plugins = []
+        self._value_factory = value_factory
+
+    def __set_name__(self, owner: type[Any], name: str) -> None:
+        # @todo Do we use the service name anywhere? We should once we write a little API to discover a service provder's
+        # @todo plugin services...
```
bartfeenstra: Expose the service name using a `name` property

betty/service/plugin/service.py
```diff
@@ -87,43 +114,104 @@ def keys(self) -> Sequence[MachineName]:
...
+    @overload
+    def __get__[ServiceProviderT](
+        self, instance: ServiceProviderT, owner: type[ServiceProviderT]
+    ) -> ValueT:
+        pass
+
+    def __get__(self, instance, owner):
+        if instance is None:
+            return self
+        return getattr(instance, self._attr_name)
```
bartfeenstra: This errors (and will continue to) if the service provider has not bootstrapped yet. We should probably require the owner to also be a life cycle so we can assert it has been bootstrapped.

betty/service/plugin/service.py
```diff
@@ -87,43 +114,104 @@ def keys(self) -> Sequence[MachineName]:
...
+    def __set_name__(self, owner: type[Any], name: str) -> None:
+        # @todo Do we use the service name anywhere? We should once we write a little API to discover a service provder's
+        # @todo plugin services...
+        self._service_name = name.strip("_")
+        self._attr_name = f"_plugin_service_{self._service_name}"
+
+    def init_with(self, instance: Any, plugins: InitT, /) -> None:
+        getattrdefault(instance, "_init_plugins", list).extend(plugins)
+
+    def _init(self, instance: Any, items: Iterable[Iterable[ItemT]], /) -> None:
```
bartfeenstra: This should accept a service level and an iterable of iterables of **plugin definitions**. Individual service types can then decide how to handle them themselves, e.g. whether to upcast the definitions (oh, they can be manufacturers too!) to awaitable factories.

betty/service/plugin/service.py
```diff
@@ -6,43 +6,70 @@
...
+
+type SupportPlugins = ResolvablePluginDefinition | Iterable[ResolvablePluginDefinition]
+
+
 @final
-class ServicePluginCollection[
-    PluginDefinitionT: ServicePluginDefinition = ServicePluginDefinition,
-    PluginT: Plugin = Plugin,
-](KeyedCollection[MachineName, ResolvablePluginId[PluginDefinitionT], PluginT]):
+class UniquePluginCollection[PluginDefinitionT: PluginDefinition, PluginT: Plugin](
```
bartfeenstra: Make private

betty/service/plugin/service.py
```diff
@@ -87,122 +116,178 @@ def keys(self) -> Sequence[MachineName]:
...
+
+
+# @todo Refactor this once we've got everything working
+@final
+class UniquePluginService[
+    PluginDefinitionT: PluginClsDefinition,
+    PluginT: Plugin[PluginClsDefinition],
+](
+    _PluginService[
+        PluginDefinitionT, UniquePluginCollection[PluginDefinitionT, PluginT]
```
bartfeenstra: Type the return value using `KeyedCollection`

bartfeenstra: Or maybe use an alias for easy reuse?

betty/service/plugin/service.py
```diff
@@ -87,122 +116,178 @@ def keys(self) -> Sequence[MachineName]:
...
+
+
+@final
+class PluginManager(ManagedLifeCycle):
     """
     The service plugin manager.
     """
 
-    def __init__[ServicePluginTypesT: ServicePluginDefinition](
+    # @todo Add back supported plugin functionality
```
bartfeenstra: Also find a more descriptive name for this

betty/service/plugin/service.py
```diff
@@ -87,122 +119,219 @@ def keys(self) -> Sequence[MachineName]:
...
-                if service_plugin_type not in self._service_plugin_types:
-                    raise UnmetRequirement(
-                        f"{fully_qualified_name(service_plugin_type)} is not a service plugin type on {fully_qualified_name(type(self._services))}."
-                    )
-                await self._expand_requires(
-                    sorter, service_plugin_type, service_plugin_id, True
-                )
-        for supported_plugin_type in self._support_plugins:
-            for supported_plugin_id in self._support_plugins[supported_plugin_type]:
+        service_plugins: _ServicePluginGraph = defaultdict(
```
bartfeenstra: We don't even need to keep a graph. We used to, because we used to initialize and bootstrap plugins in one fell swoop, but now we are working towards lazy initialization, none of that is needed anymore. All we need to do is automatically enable required plugins, done.

bartfeenstra: That means the plugin manager really does not need to do a lot. 

Once the requirements have been collected, all the plugin manager must to is call `PluginService.init(services, some_way_of_passing_additional_plugins)`. Each service can then handle is own data model and optional lazy initialization.

betty/service/plugin/service.py
```diff
@@ -87,122 +119,219 @@ def keys(self) -> Sequence[MachineName]:
...
+    PluginService[PluginDefinitionT, Collection[Awaitable[Plugin[PluginDefinitionT]]]]
+):
+    """
+    A collection of service plugins.
+    """
+
+    def __init__(self, plugin_type: type[PluginDefinitionT], /):
+        super().__init__(plugin_type, lambda plugins: next(iter(next(iter(plugins)))))
+
+    def init_with(
```
bartfeenstra: Rename all of these to `set_init_plugins()`

bartfeenstra: Or singular.

betty/service/plugin/service.py
```diff
@@ -87,122 +119,219 @@ def keys(self) -> Sequence[MachineName]:
...
+        self, instance: Any, *plugins: ServicePlugin[PluginDefinitionT]
+    ) -> None:
+        """
+        Enable the given plugins when initializing the service.
+        """
+        self._set_init_plugins(instance, *plugins)
+
+
+type _ServicePluginGraph = MutableMapping[
```
bartfeenstra: Remove?

betty/service/plugin/service.py
bartfeenstra: Do sort plugins by ID before passing them into the services. So basically we want to pass an updated collection of plugins when calling this method.

betty/service/plugin/service.py
```diff
@@ -87,122 +119,219 @@ def keys(self) -> Sequence[MachineName]:
...
-] = Collection[type[ServicePluginT]]
-type _Plugins[ServicePluginT: PluginClsDefinition = PluginClsDefinition] = Iterable[
-    PluginManufacturer[ServicePluginT, Plugin[ServicePluginT]]
-    | type[Plugin[ServicePluginT]]
+@final
+class SinglePluginService[PluginDefinitionT: PluginClsDefinition](
+    PluginService[PluginDefinitionT, Awaitable[Plugin[PluginDefinitionT]]]
+):
+    """
+    A single service plugin.
```
bartfeenstra: Single plugin service

betty/service/plugin/service.py
```diff
@@ -87,122 +119,219 @@ def keys(self) -> Sequence[MachineName]:
...
+        self, instance: Any, plugin: ServicePlugin[PluginDefinitionT], /
+    ) -> None:
+        """
+        Enable the given plugin when initializing the service.
+        """
+        self._set_init_plugins(instance, plugin)
+
+
+@final
+class PluginCollectionService[PluginDefinitionT: PluginClsDefinition](
```
bartfeenstra: Do we use this class for anything at all?

betty/service/plugin/service.py
```diff
@@ -87,122 +119,219 @@ def keys(self) -> Sequence[MachineName]:
...
+                    service,
+                    init_plugin.plugin_id
+                    if isinstance(init_plugin, PluginManufacturer)
+                    else init_plugin.plugin().id,
+                    True,
                 )
-        for service_plugin_type in self._service_plugin_types:
-            async for plugin in self._services.plugins[service_plugin_type]:
+            async for plugin in self._services.plugins[service.plugin_type]:
                 if plugin.auto:
```
bartfeenstra: Should the services decide whether or not they want auto plugins to be initialized?

betty/service/plugin/service.py
```diff
@@ -87,122 +119,219 @@ def keys(self) -> Sequence[MachineName]:
...
-                await self._expand_requires(
-                    sorter, service_plugin_type, service_plugin_id, True
-                )
-        for supported_plugin_type in self._support_plugins:
-            for supported_plugin_id in self._support_plugins[supported_plugin_type]:
+        service_plugins: _ServicePluginGraph = defaultdict(
+            lambda: defaultdict(lambda: defaultdict(list))
+        )
+        for service in self._plugin_services:
+            for init_plugin in service.get_init_plugins(self._service_provider):
```
bartfeenstra: If we pass on a service level here, plugin services can add auto plugins themselves if they want to.

betty/service/plugin/service.py
```diff
@@ -87,122 +119,219 @@ def keys(self) -> Sequence[MachineName]:
...
+    def cls(self) -> type[Any]:
+        """
+        The class this service is located on.
+        """
+        return self._cls
 
+    @final
     @property
-    def auto(self) -> bool:
+    def name(self) -> str:
```
bartfeenstra: Do we even use the service name anywhere?

betty/service/plugin/service.py
```diff
@@ -87,122 +119,219 @@ def keys(self) -> Sequence[MachineName]:
...
+    @final
+    def get(self, instance: Any) -> ServiceValueT:
+        """
+        Get the plugins for the given instance.
+        """
+        return getattr(instance, self._instance_value_attr_name)
+
+
+@final
+class UniquePluginService[
```
bartfeenstra: Add a variant for plugin definitions as well, which we can use to make CSS and JS resource plugins classless.

betty/service/plugin/service/keyed.py
```diff
@@ -0,0 +1,65 @@
...
+        services: ServiceLevel,
+        plugins: Sequence[ServicePlugin[PluginDefinitionT]],
+        /,
+    ) -> KeyedCollection[
+        MachineName, ResolvablePluginId[PluginDefinitionT], Awaitable[PluginT]
+    ]:
+        return KeyedCollectionAdapter(
+            {
+                resolve_service_plugin_id(plugin): MultiAwaitable(
+                    lambda plugin=plugin: services.factory.new(plugin)
```
bartfeenstra: Here as well as for single plugin services, plugins must be bootstrapped, and shut down when the service provider shuts down. 

Can we make a helper that handles the factory and bootstrap for us and that we can inject into `MultiAwaitable`?

betty/project/__init__.py
```diff
@@ -127,6 +115,19 @@ class Project(ChainedServiceLevel[App], ServicePluginProvider):
          - :py:class:`betty.project.data.ProjectConfiguration`
     """
 
+    _assets = KeyedPluginsService(AssetDefinition)
```
bartfeenstra: This should be public. How can we do that without conflicting with the `assets` service property?

betty/service/plugin/service.py
```diff
@@ -1,320 +1,570 @@
...
+) -> Manufacturer[PluginT]:
+    """
+    Resolve a service plugin definition to its plugin manufacturer.
+    """
+    if isinstance(plugin, PluginClsDefinition):
+        return plugin.cls  # ty:ignore[invalid-return-type]
+    return plugin  # ty:ignore[invalid-return-type]
+
+
+def _new_awaitable_service_plugin[
```
bartfeenstra: Inline this as we only ever call it once

betty/service/plugin/service.py
```diff
@@ -1,320 +1,570 @@
...
+    def get(self, instance: Any) -> ValueT:
+        """
+        Get the plugins for the given instance.
+        """
+        return getattr(instance, self.__instance_value_attr_name)
 
 
 @final
-class ServicePluginManager(ManagedLifeCycle):
+class PluginManager(ManagedLifeCycle):
```
bartfeenstra: Rename to `PluginServiceInitializer`

betty/service/plugin/service.py
```diff
@@ -1,320 +1,570 @@
...
+    @final
+    @override
+    async def _init_value_item(
         self,
-        plugin_type: type[
-            Intersection[ServicePluginDefinitionT, PluginDefinition[PluginT]]
-        ]
-        | str,
+        services: ServiceLevel,
+        instance: ManagedLifeCycle,
```
bartfeenstra: Don't use the instance but the plugin manager/initializer, because we know for a fact that that is a managed life cycle.

bartfeenstra: Unless we ensure that the instance is a managed life cycle when initializing the plugin manager/initializer

betty/service/plugin/service.py
```diff
@@ -1,320 +1,570 @@
...
+        ],
+    ],
+):  # ty:ignore[invalid-generic-class]
+    """
+    A service of plugins keyed by their IDs.
+    """
+
+
+class _PluginServiceRequirement[PluginDefinitionT: PluginDefinition]:
+    def __init__(
```
bartfeenstra: Can we overload the intializer to add support for plugin instances while keeping the class generic over `PluginDefinition`? Then we may be able to use a single class for both instance and definition services.

betty/project/__init__.py
bartfeenstra: This is a problem that may occur elsewhere, now and in the future. Can we allow `PluginService` (and all subclasses, including the final ones) to accept an optional value factory? We can use that to replace `Project.assets` with a callback in the `Project._assets` definition and rename that to `Project.assets`

bartfeenstra: If we also add an API method to `PluginService` to get all current values (or at least the current values' plugin definitions), we can use that to check plugin service requirements runtime.

betty/service/plugin/service.py
```diff
@@ -1,320 +1,577 @@
...
-        super().__init__(plugin_id, *args, **kwargs)
+    _cls: type[Any]
+    _name: str
+
+    def __init__(self, plugin_type: type[PluginDefinitionT], *, auto: bool = True):
+        self._plugin_type = plugin_type
         self._auto = auto
 
+    @final
+    def __set_name__(self, owner: type[Any], name: str) -> None:
```
bartfeenstra: Assert that no name was set yet

betty/service/plugin/service/__init__.py
```diff
@@ -1,320 +1,577 @@
...
+    def get_init_plugins(self, instance: Any, /) -> Sequence[InitT]:
+        """
+        Get the initial plugins for the given instance.
+        """
+        # Create a new iterable so we can add new init plugins while iterating over the existing ones.
+        return tuple(self.__get_init_plugins(instance))
+
+    # @todo Make this and other operations concurrency-safe.
+    @final
+    def add_init_plugins(self, instance: Any, *plugins: InitT) -> None:
```
bartfeenstra, line 121: Assert that the service has not yet been initialized

betty/service/plugin/service.py
```diff
@@ -1,320 +1,577 @@
...
+    # @todo Make this and other operations concurrency-safe.
+    @final
+    def add_init_plugins(self, instance: Any, *plugins: InitT) -> None:
+        """
+        Add one or more plugins to initialize.
+        """
+        self.__get_init_plugins(instance).extend(plugins)
+
+    @final
+    async def init(
```
bartfeenstra: Assert that the service has not yet been initialized

betty/service/plugin/service.py
```diff
@@ -1,315 +1,591 @@
...
-    def __init__(self, plugins: Iterable[Iterable[PluginT]], /):
-        self._batches = tuple(map(tuple, plugins))
-        self._all = {
-            plugin.plugin().id: plugin for batch in self._batches for plugin in batch
-        }
 
-    @override
-    def __len__(self) -> int:
-        return len(self._all)
+type ServicePluginInstance[PluginDefinitionT: PluginClsDefinition, PluginT: Plugin] = (
```
bartfeenstra: Can we remove PluginT and just do `Plugin[PluginDefinitionT]`?

bartfeenstra: Let's get rid of some intersections if we can

betty/service/plugin/service.py
```diff
@@ -1,315 +1,591 @@
...
-        *args: Any,
-        auto: bool = False,
-        **kwargs: Any,
-    ):
-        super().__init__(plugin_id, *args, **kwargs)
+    # @todo Accept a modifier callable
+    # @todo
+    def __init__(self, plugin_type: type[PluginDefinitionT], *, auto: bool = True):
+        super().__init__(self._new_service)
+        self._plugin_type = plugin_type
```
bartfeenstra: Make private

betty/app/__init__.py
```diff
@@ -80,17 +85,21 @@ class App(
...
+        assets: ServicePluginInstances[AssetDefinition, Asset] = (),
         cache_directory: Path | None = None,
         cache: TypedSynchronousServiceOrFactory[App, Cache[Any]] | None = None,
         locale: ResolvableLocale | None = None,
         plugins: Plugins | None = None,
         process_pool: TypedSynchronousServiceOrFactory[App, futures.ProcessPoolExecutor]
         | None = None,
-        service_plugins: ServicePlugins[AppServicePlugin] = (),
-        support_plugins: SupportPlugins = (),
+        rate_limits: ServicePluginInstances[RateLimitDefinition, RateLimit] = (),
```
bartfeenstra: Can we remove the plugin class type var from these type aliases?

betty/service/plugin/service.py
```diff
@@ -1,315 +1,639 @@
...
+            },
+            key_resolver=resolve_plugin_id,
         )
+
+        return KeyedCollectionAdapter(
+            # @todo If the plugin type is OrderedPluginDefinition, topologically sort the plugins as well.
+            sorted(
+                (
+                    (
+                        resolve_service_plugin_id(plugin),
```
bartfeenstra: Can we make these resolvers private? Their typing is highly specific to our 4 concrete plugin service classes in this file.

betty/service/plugin/service/__init__.py
```diff
@@ -1,315 +1,639 @@
...
+        if instance is None:
+            return self
+        return partial(PluginServiceRequirement, instance)
+
+
+class PluginServiceManager[
+    ServiceProviderT: PluginServiceProvider,
+    PluginDefinitionT: PluginDefinition,
+    GetServiceT,
+    InitT,
```
bartfeenstra, line 49: `InitT` MUST be a subtype of a resolvable plugin ID. Or do we need to replace the `resolve_*()` functions with methods that take `InitT` and return things like IDs?

betty/service/plugin/service.py
```diff
@@ -1,315 +1,639 @@
...
+        """
+        Add one or more plugins to initialize.
+        """
+        init_plugins = self.__get_init_plugins(instance)
+        for plugin in plugins:
+            plugin_id = resolve_service_plugin_id(plugin)
+            if not isinstance(init_plugins.get(plugin_id, None), PluginManufacturer):
+                init_plugins[plugin_id] = plugin
+            elif isinstance(plugin, PluginManufacturer):
+                raise ValueError(
```
bartfeenstra: Move this check to `.init_plugins()` implementations as it is specific to instance services only.

betty/service/plugin/service.py
```diff
@@ -1,315 +1,639 @@
...
+                    f"Cannot add more than one plugin manufacturer for {plugin_id} to service {self.id}."
+                )
+    # @todo FInish this
+    async def init_plugins(self, *plugins:InitT)->None:
```
bartfeenstra: Add the instance as an arg

betty/service/plugin/service/__init__.py
```diff
@@ -1,315 +1,639 @@
...
+        return plugin.cls  # ty:ignore[invalid-return-type]
+    return plugin
+
+
+type _MultiplePluginServiceValue[PluginDefinitionT: PluginDefinition, ValueItemT] = (
+    KeyedCollection[MachineName, ResolvablePluginId[PluginDefinitionT], ValueItemT]
+)
+
+
+class _SinglePluginServiceManager[
```
bartfeenstra: Make public. Add a base class for multiple values as well. 

Reorganize these mixins as well as the four concrete classes into submodules.

betty/service/plugin/service.py
```diff
@@ -1,315 +1,686 @@
...
 
-    @override
-    def __contains__(self, key: Any) -> bool:
-        try:
-            return resolve_plugin_id(key) in self._all
-        except ValueError:
-            return False
 
+@final
+class _PluginServiceRequirementGetter(Singleton):
```
bartfeenstra: Make public. Move all requirements code to a submodule.

betty/service/plugin/service.py
```diff
@@ -1,315 +1,686 @@
...
+
+    require: Final[_PluginServiceRequirementGetter] = _PluginServiceRequirementGetter()
+
+    def __init__(self, plugin_type: type[PluginDefinitionT], *, auto: bool = False):
+        super().__init__(self._new_service)
+        self.__plugin_type = plugin_type
+        self.__auto = auto
+
+    @final
+    def __set_name__(self, owner: type[ServiceProviderT], name: str) -> None:
```
bartfeenstra: Remove this. Make the parent final yet again. Compute the atte name dynamically instead.

bartfeenstra: Consider doing the same for the parent

betty/service/provider.py
```diff
@@ -33,10 +33,19 @@ class ServiceProvider:
...
-        super(*args, **kwargs)
+        super().__init__(*args, **kwargs)
+        self.__services = services
         for _, member in getmembers(type(self)):
             if isinstance(member, ServiceManager):
-                member.init(services, self)
+                member.init(self)
+
+    @final
+    @property
```
bartfeenstra, line 32: Extract these changes into a separate PR. Also reorganize the concrete subclasses into submodules.

betty/service/provider.py
```diff
@@ -74,10 +83,35 @@ def __init__(
         self.__service_or_factory = factory
 
     def __set_name__(self, owner: type[ServiceProviderT], name: str) -> None:
+        self.__service_cls = owner
```
bartfeenstra: No need for prefixes when using dunder private attrs

betty/service/provider.py
```diff
@@ -258,7 +289,13 @@ def _get_service(self, service: Callable[[], ServiceT], /) -> ServiceT:
         return service()
 
 
-class ServiceInitializedError(ServiceError):
+class ServiceNotYetInitialized(ServiceError):
```
bartfeenstra, line 189: Add an assertion for this

betty/service/plugin/service/__init__.py
```diff
@@ -1,316 +1,689 @@
...
+    ) -> Iterable[tuple[PluginServiceManager, PluginDefinition]]:
+        plugin = await self._services.plugins[service.plugin_type][plugin_id]
+        return (
+            (service, plugin),
+            *chain(*await gather(*map(self._collect_requirement, plugin.requires))),
+        )
+
+    async def _collect_auto_plugins(
+        self,
+    ) -> Iterable[tuple[PluginServiceManager, PluginDefinition]]:
```
bartfeenstra, line 255: Add an initializer kwarg to disable auto plugins. Add this kwarg to all plugin service providers. This is essential for test isolation.

betty/service/plugin/service.py
```diff
@@ -1,316 +1,689 @@
...
+        GetServiceT,
+    ]
+):
+    """
+    A plugin service manager.
+    """
+
+    require: Final[_PluginServiceRequirementGetter] = _PluginServiceRequirementGetter()
+
+    def __init__(self, plugin_type: type[PluginDefinitionT], *, auto: bool = False):
```
bartfeenstra: Change the default for auto to True. Update all service definitions accordingly, e.g. remove auto=True everywhere.

betty/service/plugin/service.py
```diff
@@ -1,316 +1,689 @@
...
-        **kwargs: Any,
-    ):
-        super().__init__(plugin_id, *args, **kwargs)
-        self._auto = auto
+    @final
+    def get_init_plugins(self, instance: Any, /) -> Sequence[InitT]:
+        """
+        Get the initial plugins for the given instance.
+        """
+        # Create a new iterable so we can add new init plugins while iterating over the existing ones.
```
bartfeenstra: This is no longer needed now we have init_plugins()

betty/service/plugin/service.py
```diff
@@ -1,316 +1,689 @@
...
+            for member in getmembers(type(service_provider))
+            if isinstance(member[1], PluginServiceManager)
+        )
+        self.life_cycle.on_bootstrap(self._initialize_plugin_services)
+
+    async def _initialize_plugin_services(self) -> None:
+        init_plugins = chain(
+            *await gather(
+                *[
+                    # Expand the service's own init plugins.
```
bartfeenstra: r/expand/collect in all comments

betty/service/plugin/service/__init__.py
bartfeenstra: Although we use auto plugins sparingly, and for core functionality only, that would otherwise have been hardcoded. And no tests have failed on this yet. So maybe not.

betty/service/plugin/service/instance/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+        GetServiceT,
+        ServicePluginInstance[PluginDefinitionT],
+    ]
+):
+    """
+    A service containing plugin instances.
+    """
+
+    @final
+    def _new_plugin_instance_service_item(
```
bartfeenstra, line 51: Add a new `_PluginServiceManager(PluginServiceManager)` that declares `_new_service_item()` and refactor this method into an override of that one. The goal is for our four concrete service classes not to need any method overriiof their own.

betty/service/plugin/service/instance/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+from betty.plugin.cls import Plugin, PluginClsDefinition
+from betty.plugin.factory import PluginManufacturer
+from betty.plugin.resolve import ResolvablePluginDefinition, resolve_plugin_id
+from betty.service import ServiceError
+from betty.service.plugin.service import PluginServiceManager, PluginServiceProvider
+
+if TYPE_CHECKING:
+    from betty.machine_name import MachineName
+
+type ServicePluginInstance[PluginDefinitionT: PluginClsDefinition] = (
```
bartfeenstra, line 22: Rename all the type aliases consistently across this PR

betty/service/plugin/service/__init__.py
```diff
@@ -0,0 +1,315 @@
...
+    async def _init_plugins(
+        self, instance: ServiceProviderT, *plugins: InitT
+    ) -> Iterable[InitT]:
+        """
+        Initialize the plugins.
+
+        Perform actions such as validation or ordering here.
+        """
+        # Deduplicate init plugins, where later ones override earlier ones.
+        return {
```
bartfeenstra, line 149: Move this implementation to the to-be-added `_PluginServiceManager`

betty/service/plugin/service/__init__.py
```diff
@@ -0,0 +1,315 @@
...
+        if not hasattr(instance, f"_plugin_service_plugins_{self.name}"):
+            raise ServiceNotYetInitialized(
+                f"Service {self.id}'s plugins were not yet initialized."
+            )
+
+    @abstractmethod
+    def _new_service(self, instance: ServiceProviderT, /) -> GetServiceT:
+        pass
+
+    # @todo Make this abstract and move the current default implementatin to a to-be-created
```
bartfeenstra, line 178: Fix this

betty/service/plugin/service/collection/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+                self.resolve_init_plugin_id(plugin): self._new_service_item(
+                    instance, plugin
+                )
+                for plugin in self.get_plugins(instance)
+            },
+            key_resolver=resolve_plugin_id,
+        )
+
+    @abstractmethod
+    def _new_service_item(
```
bartfeenstra, line 74: Remove this in favor of the new `_PluginServiceManager._new_service_item()`

betty/service/plugin/service/multiple.py
```diff
@@ -0,0 +1,105 @@
...
+                    sorter.add(before, plugin.id)
+            return sorter.static_order()
+        return (plugin.id for plugin in plugins)
+
+    @final
+    @override
+    def _new_service(
+        self, instance: ServiceProviderT, /
+    ) -> MultiplePluginServiceValue[PluginDefinitionT, GetServiceItemT]:
+        return KeyedCollectionAdapter(
```
bartfeenstra: Extract this into a new `KeyedCollectionPluginServiceManager(MultiplePluginServiceManager)`

betty/service/plugin/service/__init__.py
bartfeenstra: Let's not, because this implementation reflects the underlying constraint of unique/identifiable plugins in plugin services.

betty/service/plugin/service/requirement.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+
+    @overload
+    def __get__(self, instance: None, owner: type[PluginServiceManager]) -> Self:
+        pass
+
+    @overload
+    def __get__(
+        self,
+        instance: PluginServiceManager,
```
bartfeenstra: For now we only support plugin services on service levels, so type `instance` to enforce that.

betty/service/plugin/service/collection/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+                    *map(
+                        service_provider.services.plugins[self.plugin_type].get,
+                        plugins_by_id.keys(),
+                    )
+                ),
+            )
+        )
+
+    @final
+    def __sort_plugins(
```
bartfeenstra, line 58: Make this an API function next to `OrderedDefinition`. Do we sort plugins anywhere else? Make it use this new function too.

bartfeenstra: Do we have any old unused cyclic dependency errors we can repurpose?

PR #4035, opened 2026-04-03
Improve the DX of declaring requirements on service levels and service plugins

bartfeenstra:


betty/app/__init__.py
```diff
@@ -71,6 +80,9 @@ class App(
          - :py:class:`betty.app.data.AppConfiguration`
     """
 
+    requirement: Final[ServiceLevelRequirement[Self]] = _ServiceLevelRequirement[Self]()
```
bartfeenstra: Could we not put this on `ServiceLevel` and let it be inherited? We could but then we'd need a special descriptor that returns the requirement in `__get__()` based on the owning class.

bartfeenstra: In which case wed should not expose a requirement attribute directly

betty/service/level/requirement.py
```diff
@@ -0,0 +1,48 @@
...
+from __future__ import annotations
+
+from typing import final
+
+from betty.importlib import fully_qualified_name
+from betty.requirement import UnmetRequirement
+from betty.service.level import ChainedServiceLevel, ServiceLevel
+
+
+class _ServiceLevelRequirement[ServiceLevelT: ServiceLevel]:
```
bartfeenstra: Do we use this?

betty/app/__init__.py
bartfeenstra: And move this to `RequireableServiceLevel` in a submodule to prevent circular imports

betty/jinja/filter.py
```diff
@@ -33,6 +37,14 @@ class JinjaFilterDefinition(ServicePluginDefinition[JinjaFilter]):
     .. plugin_type:: jinja-filter.
     """
 
+    def __init__(
```
bartfeenstra, line 40: Extract these definition improvements into a separate PR

betty/requirement.py
bartfeenstra: Refactor this into a new mixin base class that any requirement could subclass.

bartfeenstra: Actually keeping this as a separate class/function means it can also be combined with non-classed requirements (because a requirement is just a callable).

bartfeenstra: So keep this separate but also add a mixin class.

bartfeenstra: If we let the decorator's `__get__()` method return a new callable depending on whether it is accessed through the class or instance, we don't need this buggy magic anymore.

bartfeenstra: We may need separate decorators for functions and for methods.

bartfeenstra: Can we add a decorator base class to `betty.functools` that provides this functionality? We'd save some complexity here.

bartfeenstra: Can the base class support callable objects as well?

bartfeenstra: Support decorating descriptors that return callables.

bartfeenstra: Look up why `partialmethod()` is separate and not part of `partial()`. Should we use two separate decorators for the same reason? Or does `partialmethod()` exist only because adding method support to `partial()` would have been a BC break? 

Our decorator should be able to decorate `@staticmethod` as well.

bartfeenstra: `partialmethod()` was added so as not to alter `partial()`'s existing behavior. We can just put everything in a single decorator just fine.

bartfeenstra: Add a `SupportsDunderGet[DunderGetT]` protocol that declares `__get__(...) -> DunderGetT` and use it to type the decorated item if it is a descriptor returning the decorated callable.

PR #4046, opened 2026-04-08
Make service managers reusable

bartfeenstra:


betty/asyncio.py
```diff
@@ -18,3 +20,31 @@ async def resolve_await[T](value: Awaitable[T] | T) -> T:
     if isawaitable(value):
         return await value  # ty:ignore[invalid-return-type]
     return value
+
+
+@final
+class ReAwaitable[ValueT]:
```
bartfeenstra: Remove this if we don't end up using it

betty/service/provider.py
```diff
@@ -83,8 +74,8 @@ def service(factory=None):
 
     def _service(factory):  # noqa: ANN001, ANN202
```
bartfeenstra: Merge this local function into `_ServiceDecorator`, make it a final class. With the existing overloads, that means we can remove these Ruff ignores here.

betty/service/provider.py
```diff
@@ -42,27 +33,27 @@ class _ServiceDecorator[ServiceProviderT: ServiceProvider, ServiceT](Protocol):
...
-    ) -> _AsynchronousServiceManager[ServiceProviderT, ServiceT]:
+    ) -> AsynchronousServiceManager[ServiceProviderT, ServiceT]:
         pass
 
 
 @overload
 def service[ServiceProviderT: ServiceProvider, ServiceT](
     factory: Callable[[ServiceProviderT], Awaitable[ServiceT]], /
-) -> _AsynchronousServiceManager[ServiceProviderT, ServiceT]:
+) -> AsynchronousServiceManager[ServiceProviderT, ServiceT]:
```
bartfeenstra: Type the return value using `ServiceManager`

betty/service/provider.py
```diff
@@ -42,27 +33,27 @@ class _ServiceDecorator[ServiceProviderT: ServiceProvider, ServiceT](Protocol):
...
     pass
 
 
 @overload
 def service[ServiceProviderT: ServiceProvider, ServiceT](
     factory: Callable[[ServiceProviderT], ServiceT], /
-) -> _SynchronousServiceManager[ServiceProviderT, ServiceT]:
+) -> SynchronousServiceManager[ServiceProviderT, ServiceT]:
```
bartfeenstra: Type the return value using `ServiceManager`

betty/service/provider.py
```diff
@@ -202,12 +197,23 @@ def override_factory(
...
+        self, factory: ServiceFactory[ServiceProviderT, Awaitable[ServiceT]], /
+    ):
+        super().__init__(factory)
+        self._lock = AsynchronizedLock.new_threadsafe()
+
+    @final
     @override
     async def _get(self, instance: ServiceProviderT, /) -> ServiceT:
-        async with instance._ledger.ledger(self.name):
+        async with self._lock:
```
bartfeenstra: If the value is a life cycle, bootstrap it and attach it to the service provider's life cycle manager.

betty/service/provider.py
```diff
@@ -186,6 +180,7 @@ def override(self, instance: ServiceProviderT, service: ServiceT, /) -> None:
         setattr(instance, self._service_attr_name, service)
         setattr(instance, self._service_override_attr_name, True)
 
+    @final
     def override_factory(
```
bartfeenstra: Do we use this anywhere, and if so, should we allow awaitable factories for asynchronous services?

betty/asyncio.py
bartfeenstra: Can we make `ReAwaitable(Awaitable)` a type and refactor the class into `LazyReAwaitable`, and use the class solely to indicate any type that can be awaited more than once? We can use `ReAwaitable` to type asynchronous services' return values, indicating that such services can be injected safely and awaited only when actually needed. We may also want to add types like `type OptionalAwaitable[T] = Awaitable[T] | T` and  `type OptionalReAwaitable[T] = ReAwaitable[T] | T`

betty/service/provider.py
```diff
@@ -199,38 +216,88 @@ def override_factory(
...
+            self.__service_attr_name,
+            LazyReAwaitable(
+                lambda: self.__init(instance), AsynchronizedLock.new_threadsafe()
+            ),
+        )
 
-class _SynchronousServiceManager[ServiceProviderT: ServiceProvider, ServiceT](
-    ServiceManager[ServiceProviderT, ServiceT, ServiceT]
+    # @todo For both sync and async services, support the abstract concepts of:
+    # @todo - The (unwrapped/resolved) service type: ServiceValueT
```
bartfeenstra: If we merge `.override()` and `.override_factory()` and require overrides to be of `LazyServiceT` (easy using lambdas and `LazyReAwaitable`), we can greatly simplify our logic and generics.

bartfeenstra: Do we want a `LazyReCallable` as a synchronous equivalent to `LazyReAwaitable`?

bartfeenstra: Then again merging those two methods would make overriding services with values more verbose.

betty/functools.py
```diff
@@ -277,3 +278,33 @@ def __call__(self, decorated=None):
...
+    def __init__(self, producer: Callable[[], ValueT], lock: threading.Lock, /):
+        self._producer = producer
+        self._called = False
+        self._lock = lock
+
+    def __call__(self) -> ValueT:
+        """
+        Get the value.
+        """
+        if not self._called:
```
bartfeenstra: Add a comment explaining that we check the flag before and after acquiring the lock to prevent lock acquisitions once the value has been created.

betty/asyncio.py
```diff
@@ -4,17 +4,66 @@
...
+    _value: ValueT
+
+    def __init__(self, producer: Callable[[], Awaitable[ValueT]], lock: Lock, /):
+        self._producer = producer
+        self._awaited = False
+        self._lock = lock
+
+    @override
+    def __await__(self) -> Generator[Any, Any, ValueT]:
+        if not self._awaited:
```
bartfeenstra: Add a comment explaining that we check the flag before and after acquiring the lock to prevent lock acquisitions once the value has been created.

betty/asyncio.py
```diff
@@ -4,17 +4,66 @@
...
+
+    The proxied awaitable will at most be awaited once.
+    """
+
+    __slots__ = "_awaited", "_lock", "_producer", "_value"
+
+    _value: ValueT
+
+    def __init__(self, producer: Callable[[], Awaitable[ValueT]], lock: Lock, /):
+        self._producer = producer
```
bartfeenstra: Rename to `factory`

betty/functools.py
```diff
@@ -277,3 +278,33 @@ def __call__(self, decorated=None):
...
+    A value that can be called multiple times while always returning the exact same value.
+
+    The proxied callable will at most be called once.
+    """
+
+    __slots__ = "_called", "_lock", "_producer", "_value"
+    _value: ValueT
+
+    def __init__(self, producer: Callable[[], ValueT], lock: threading.Lock, /):
+        self._producer = producer
```
bartfeenstra: Rename to `factory`

betty/functools.py
```diff
@@ -277,3 +278,33 @@ def __call__(self, decorated=None):
...
+    """
+    A value that can be called multiple times while always returning the exact same value.
+
+    The proxied callable will at most be called once.
+    """
+
+    __slots__ = "_called", "_lock", "_producer", "_value"
+    _value: ValueT
+
+    def __init__(self, producer: Callable[[], ValueT], lock: threading.Lock, /):
```
bartfeenstra: Set locks internally

betty/plugins/extension/http_api_doc/__init__.py
```diff
@@ -30,14 +33,18 @@
...
-class HttpApiDoc(EntryPointProvider):
+class HttpApiDoc(EntryPointProvider, Manufacturable):
     """
     .. plugin:: extension:http-api-doc.
     """
 
-    # Provide an initializer without arguments so the factory can call it.
-    def __init__(self):
-        super().__init__()
+    def __init__(self, *, services: ServiceLevel):
```
bartfeenstra: If we remove `ServiceProvider` as a superclass from `Extension` for now, we can undo these changes on most extensions in this PR.

betty/service/provider.py
```diff
@@ -182,14 +189,14 @@ def override(self, instance: ServiceProviderT, service: ServiceT, /) -> None:
 
         This MUST only be called from ``instance.__init__()``.
         """
-        self._assert_not_initialized(instance)
-        setattr(instance, self._service_attr_name, service)
-        setattr(instance, self._service_override_attr_name, True)
+        self._assert_service_not_initialized(instance)
+        setattr(instance, self.__service_override_attr_name, service)
```
bartfeenstra: Remove this attribute entirely and just call `self.override_factory()`? But then how would we convert `ServiceT` to `FactoryServiceT`?

betty/service/provider.py
```diff
@@ -127,110 +110,123 @@ def __get__(self, instance: None, owner: type[ServiceProviderT], /) -> Self:
...
+    """
+    Manage an asynchronous service.
+    """
 
+    @final
+    @override
+    def _init_lazy_service(
+        self, instance: ServiceProviderT, /
+    ) -> ReAwaitable[ServiceT]:
+        return LazyReAwaitable(lambda: resolve_await(self._new_service(instance)))
```
bartfeenstra: Two things:
- Add a helper method that takes a value and synchronizes it's life cycle:
  - If the value is not a life cycle, do nothing
  - If the value is a life cycle:
    - if the service provider is not a managed life cycle, raise an error
    - if the value's and service's life cycles are in sync, do nothing
    - else, bootstrap the value and add its shutdown to the service provider's managed life cycle
- Call this helper in the lazy awaitable callback. 
- this helper can also be used by future multiple-plugin services

bartfeenstra: Maybe this helper should be a generic global function somewhere.

betty/app/__init__.py
```diff
@@ -79,18 +79,40 @@ def __init__(
         self,
         *,
         cache_directory: Path | None = None,
-        cache_factory: ServiceFactory[App, Cache[Any]] | None = None,
+        cache: Cache[Any] | Callable[[App], Cache[Any]] | None = None,
```
bartfeenstra: The callable may return an awaitable as well. Use type aliases for these.

bartfeenstra: Final class `Service`
Types:
- `SychronousServiceFactory`
- `SynchronousServiceOrFactory`
- `AsychronousServiceFactory`
- `AsynchronousServiceOrFactory`

PR #4075, opened 2026-04-15
Improve file I/O performance

### LINKED ISSUE #4074: Consider replacing aiofiles
[Online benchmarks](https://github.com/mosquito/aiofile/issues/) appear to indicate it is significantly slower than using `asyncio.to_thread()`. Explore call sites and make an overview of the different ways aiofiles is used. Can we provide fast helpers for common use cases? I believe `async with aiofiles.open() as f: await f.read()` may dispatch both the open and read to a thread separately. Making a synchronous helper for these two, and dispatching the helper to a thread once, may be faster.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/4074

## Follow-up
- Reduce `mkdir()` calls. Especially in generation jobs where we have been able to declare dependencies for a while, we can use that to ensure the necessary directories were created instead of having thousands of needless `mkdir(exist_ok=True)` calls


bartfeenstra:
The test for posterity:
```python
from asyncio import gather, run, to_thread
from pathlib import Path
from tempfile import TemporaryDirectory
from time import process_time_ns

import aiofiles

MODE = "w"
CONTENT = "Hello, world!"


async def to_thread_read(file: Path, /) -> None:
    raise NotImplementedError


def _to_thread_write(file: Path) -> None:
    with open(file, mode=MODE) as f:
        f.write(CONTENT)


async def to_thread_write(file: Path, /) -> None:
    await to_thread(_to_thread_write, file)


async def aiofiles_read(file: Path, /) -> None:
    raise NotImplementedError


async def aiofiles_write(file: Path, /) -> None:
    async with aiofiles.open(file, mode=MODE) as f:
        await f.write(CONTENT)


TESTS = (
    # @todo
    # to_thread_read,
    to_thread_write,
    # aiofiles_read,
    aiofiles_write,
)

ITERATIONS = 9999


async def main() -> None:
    with TemporaryDirectory() as temporary_directory:
        for test in TESTS:
            test_directory = Path(temporary_directory) / test.__name__
            test_directory.mkdir()
            start = process_time_ns()
            await gather(*[
                test(test_directory / f"file_{iteration}")
                for iteration in range(ITERATIONS)
            ])
            end = process_time_ns()
            end - start


if __name__ == "__main__":
    run(main())
```

PR #4077, opened 2026-04-15
Improve `AsynchronizedLock`'s CPU performance

### LINKED ISSUE #4076: Benchmark `AsynchronizedLock`

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/4076

### DIFF
--- a/betty/concurrent.py
+++ b/betty/concurrent.py
@@ -10,7 +10,7 @@ from asyncio import sleep
 from collections.abc import AsyncIterator, Hashable, MutableMapping
 from math import floor
 from types import TracebackType
-from typing import Self, final, override
+from typing import final, override
 
 from betty.typing import threadsafe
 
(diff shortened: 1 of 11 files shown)


betty/concurrent.py
```diff
@@ -66,14 +66,7 @@ def lock(self) -> threading.Lock:
...
     async def acquire(self, *, wait: bool = True) -> bool:
-        while not self._lock.acquire(blocking=False):
-            if not wait:
-                return False
-            # Sleeping for zero seconds does not actually sleep, but gives the event
-            # loop a chance to progress other tasks while we wait for another chance
-            # to acquire the acquirable.
-            await sleep(0)
-        return True
+        return await to_thread(self._lock.acquire, blocking=wait)
```
bartfeenstra: This blocks as soon as this is called concurrently more times than there are threads available in the pool.

bartfeenstra: What if we use exponential back off for now?

PR #4097, opened 2026-04-20
Improve the factories

### LINKED ISSUE #3893: Improve `Factory`
## To do
- Raise factory API specific errors
- Document that `new()` may raise this error as well as `UnmetRequirement`.
- Allow `Any` to be provided as a target, with return value `Never`?
- If the provided target cannot be initialized, raise a factory API error (unlocalized) with clear instructions.

### LINKED ISSUE #3991: Simplify `Factory`
Can we remove support for data entirely? I believe the only place that uses it is `PluginManufacturer` which we can refactor to handle new instance creation using data itself internally, and if no data handling is needed, use `Factory.new` on the plugin class. The reason this is not done currently, IIRC, is because plugin manufacturers received this ability after `Factory` was last refactored. 

At the very least, remove **portable** data support from `Factory`.

Also, can we replace typing on `type` with `Callable`?

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3991, https://github.com/bartfeenstra/betty/issues/3893


betty/factory.py
```diff
@@ -69,58 +68,33 @@ def __init__(self, services: ServiceLevel, /):
...
-                ).format(target=fully_qualified_name(target))
-            )
-        if not isinstance(data, Data):
-            data = target.new_data_cls().data().porter.load(data)
-        return await target.new(self._services, data)
+        if isinstance(target, type) and issubclass(target, Manufacturable):
+            return await target.new(self._services)
+        signature = inspect.signature(target)
+        try:
+            signature.bind()
```
bartfeenstra, line 90: Instead of using bind(), which does not appear to be very exact, check if the signature takes exactly ONE argument which we assume to be the service level. If not, check if the signature has zero required arguments. If not, raise an error. This means we can remove a lot of __init__() overrides from classes like plugins which currently only exist because the factory would call them with a service level due to inherited arguments with default values. As for that, we may actually want to treat classes separately and always call them without arguments.

bartfeenstra: `Intersection[type[T], Callable[[], T]]`

betty/factory.py
```diff
@@ -65,62 +75,56 @@ class Factory:
     The object factory.
     """
 
+    _SIGNATURE_MESSAGE = f'any required arguments, except an optional first argument typed on {fully_qualified_name(ServiceLevel)} and/or named "services".'
```
bartfeenstra: "optional" implies that the value may be None. What we mean is that that argument may or may not exist. 

However, do test that the factory works if the type hint is not EXACTLY ServiceLevel.

PR #4144, opened 2026-04-29
Use Typeshed's `StrPath`

### LINKED ISSUE #4143: Use Typeshed's `StrPath`
Use Typeshed's `StrPath` instead of `pathlib.Path` where we can, to allow developers to use strings instead of having to wrap them in `Path` everywhere.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/4143


betty/plugins/extension/webpack/jobs.py
```diff
@@ -27,9 +27,10 @@ def __init__(self, *, builder: Builder, www_directory: Path, cache_directory: Pa
     @override
     async def do(self, scheduler: Scheduler, /) -> None:
         context = scheduler.context
-        build_directory_path = await self._generate_ensure_build_directory(context)
-        context._webpack_build_directory_path = build_directory_path  # ty:ignore[unresolved-attribute]
-        await self._copy_build_directory(build_directory_path, self._www_directory)
+        build_directory = await self._generate_ensure_build_directory(context)
+        # @todo Do we use this at all anymore?
```
bartfeenstra: Address this

betty/pathlib/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+"""
+
+from os import PathLike
+from pathlib import Path
+
+# This mimics _typeshed.StrPath
+type StrPath = str | PathLike[str]
+
+
+def resolve_path(path: StrPath) -> Path:
```
bartfeenstra, line 12: Benchmark if this is actually faster than always calling `Path(...)`

bartfeenstra: Get CPU and memory benchmarks

PR #4161, opened 2026-05-02
Reduce the JSON schema API

### LINKED ISSUE #3829: Remove the JSON schema API
It's verbose and not that good. Refactor schema building so we do not need to edit other objects's schemas.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3829

## June 22 thoughts
- Maybe start from scratch? We may need to keep a version of `Schema`
- linked data schemas must produce a portable data schema and an optional definition name. Alternatively, if we can somehow get a unique name for/from each schema, we only need to produce the schema portable data.
- dumped linked data must produce the portable data and optional contexts. Alternatively, check our use cases. I think we only ever needed this to apply contexts to arrays of items, but can we instead not embed the context in each item? No, not if that item isn't itself an object...

## To do
- Fix `@todo`s
- Remove every `schema.py` 
- Ensure schemas and their definitions can be merged


betty/linked_data.py
```diff
@@ -5,25 +5,36 @@
...
+@dataclass(frozen=True)
+class LinkedDataSchema:
+    """
+    JSON Schema information for a JSON-LD dump.
+    """
+
+    schema: PortableMapping
+
+    # @todo
+    def add_context(self) -> None:
```
bartfeenstra: Remove this as it's an immutable class

betty/linked_data.py
```diff
@@ -5,25 +5,36 @@
...
 
 
+@final
+@dataclass(frozen=True)
+class LinkedDataSchema:
+    """
+    JSON Schema information for a JSON-LD dump.
+    """
+
+    schema: PortableMapping
```
bartfeenstra: Expand this with definitions

betty/linked_data.py
```diff
@@ -5,25 +5,36 @@
...
 
 if TYPE_CHECKING:
-    from betty.locale.localizable import ResolvableLocalizable
-    from betty.plugins.entity.link import Link
     from betty.project import Project
 
 
+@final
+@dataclass(frozen=True)
+class LinkedDataSchema:
```
bartfeenstra: Decouple this from linked data and move to the JSON Schema API

betty/json_schema.py
```diff
@@ -4,472 +4,99 @@
 
 from __future__ import annotations
 
-import enum
-from typing import TYPE_CHECKING, Any, ClassVar, cast, final, override
+from dataclasses import KW_ONLY, dataclass
+from typing import TYPE_CHECKING, Any, Final, final
 
 from jsonschema.validators import Draft202012Validator
```
bartfeenstra: If we move the validation code to a submodule, the main JSON Schema module won't have any importa beyond Python's stdlib, and we can safely import `Schema` from every API without risking complex or circular imports (meaning the reason we has all those schema.py files will no longer be relevant)

betty/json_schema.py
```diff
@@ -4,472 +4,99 @@
...
-
 if TYPE_CHECKING:
-    from betty.locale.localizable import ResolvableLocalizable
-
-
-class Schema:
-    """
-    A JSON Schema.
-
-    All schemas using this class **MUST** follow JSON Schema Draft 2020-12.
```
bartfeenstra, line 24: Add this comment back, but to the main module docstring.

PR #4162, opened 2026-05-02
Remove `App.entity_types`

### LINKED ISSUE #4146: Remove `App.entity_types`
Data plugin types may be in abundance, including those that are discovered dynamically. We only added this service as a quick stopgap to be able to sort entity types for their primary links. However, entity order means nothing, really. 

## However...
- Keeping services for all plugin types does mean each has a single authoritative source of truth for things like order and sorting. However, we're right that we don't want just any and all definitions included. Also, for sorting, we only ever really want to sort a limited set of values that are or represent plugins. Like we want to sort a person's events, or all events in an ancestry. The only time we ever want to sort all plugins is in `betty about` and we sort by plugin ID there.

## To do
- Remove the service 
- Make entity definitions no longer ordered
- for primary links, perhaps sort by plugin ID just to guarantee a stable link order? Links can (and perhaps should) be overriden for most projects anyway, and the overrides are always ordered. 
- for demo projects, explicitly configure the HTML entity lists

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/4146

### DIFF
--- a/betty/service_level/__init__.py
+++ b/betty/service_level/__init__.py
@@ -9,10 +9,7 @@ from importlib import metadata
 from typing import TYPE_CHECKING, Any
 
 from betty.collection.keyed.adapter import KeyedCollectionAdapter
-from betty.collection.keyed.error import ErroringKeyedCollection
 from betty.life_cycle.manage import ManagedLifeCycle
-from betty.machine_name import MachineName
-from betty.plugin import PluginDefinition
 from betty.plugin.resolve import resolve_plugin_type_id
 from betty.service import ServiceProvider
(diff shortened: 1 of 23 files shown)


betty/service_level/__init__.py
```diff
@@ -113,3 +114,12 @@ def upstream(self) -> UpstreamT:
         The upstream service level.
         """
         return self._upstream
+
+
+# @todo
+class _PluginTypeCollection(KeyedCollectionProxy[
```
bartfeenstra: Finishing this should fix our typing errors

PR #4192, opened 2026-05-06
Add `@Property.getter`, `@Property.setter`, and `@Property.default`

### LINKED ISSUE #4189: Add `Property.getter()` and `.setter()`
Like `@property`, these would be helpful for computed properties.

Similarly, `@Property` should be usable as an instance method decorator (a shortcut for `@Property().getter`)

A setter is the same as a resolver. Make sure these two are unified! In fact, remove resolver support, and fold it all into the new setter support? Do check if that works for subclasses!

Implement for `betty.plugins.entity.link.Link.label`. **CAUTION** this specific use case may not benefit from it, because other functionality within the same class must be able to access the internal label, even if that is None. This may mean the entire purpose for this PR is no longer relevant. **RECONSIDER**

We can only do this once we have a unified solution for decorating properties. Add `classtools.Wrapper` and take inspiration from `OptionalDefinition`?

## Blocked by
- https://github.com/bartfeenstra/betty/pull/4190

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/4189

## To do
- For getters and setters, if they change the value type, they MUST provide an updated data definition. However, we have no way of creating (modified) copies of data definitions. This really applies to any decorator that changes one of the types.
- Solve the `DefaultProperty` racing problem or find another approach that does not have race conditions, e.g. require an instance's `__init__()` to set a value. We could use locks but with properties being used on entities, of which thousands may exist, we may want to **hook into the instance's `__init__()` instead. Can we extract this functionality from service managers into a new descriptor API?** actually, we cannot use locks because we can't create a unique one for each instance before we have to acquire it.
  - Add getter, setter, and default support to the descriptor API? Allow the getter API to only provide getters that return the same type. Add `AnyGetterDescriptor(Descriptor)` to add getters that change the return type.
  - Add `HasDescriptors` and move `ServiceProvider`'s descriptor init code there.
  - Let's also add something like `BootstrappableDescriptor` and `HasBootstrappableDescriptors` to home the init code currently in `PluginServiceProvider`.
  - Add `GetterDescriptor`, `AnyGetterDescriptor`, `SetterDescriptor`, `DeleterDescriptor`, and `DefaultDescriptor`, all subclasses of `Descriptor` (or even `BootstrappableDescriptor`?
  - How do we handle `BootstrappableDescriptor.getter()` et al returning descriptors that are not bootstrappable? Should they be bootstrappable descriptors as well that just proxy if their upstream is bootstrappable?
- Add an API to get the underlying value? Useful for `Link.has_label` as well as testing.

### DIFF
--- a/betty/property.py
+++ b/betty/property.py
@@ -9,7 +9,6 @@ from typing import TYPE_CHECKING, Any, Self, cast, final, overload, override
 
 from betty.data import OptionalDefinition
 from betty.datas.aggregate.record.object import Attr, AttrDefinition
-from betty.functools import passthrough
 from betty.importlib import fully_qualified_name
 from betty.typing import Void
 
@@ -20,26 +19,32 @@ if TYPE_CHECKING:
     from betty.locale.localizable import ResolvableLocalizable
 
 
-class _Property[ValueGetT, ValueSetT](Attr[ValueGetT], ABC):
-    _attr_name: str
+class Property[ValueGetT, ValueSetT](Attr[ValueGetT], ABC):
+    """
+    An instance property backed by a data definition.
+    """
 
-    def __init__(
-        self,
-        attr: AttrDefinition[ValueGetT],
-        *,
-        resolver: Callable[[ValueSetT | ValueGetT], ValueGetT] = passthrough,
-    ):
-        self._attr = attr
-        self._resolver = resolver
+    __name: str
+
+    def __init__(self, attr: AttrDefinition[ValueGetT], /):
+        self.__attr = attr
 
     def __set_name__(self, owner: type[Any], name: str) -> None:
-        self._attr_name = f"_{name}"
+        self.__name = f"_{name}"
 
     @final
     @override
     @property
     def attr(self) -> AttrDefinition[ValueGetT]:
-        return self._attr
+        return self.__attr
+
+    @final
+    @property
+    def name(self) -> str:
+        """
+        The property/attribute name.
+        """
+        return self.__name
 
     @overload
     def __get__(self, instance: None, owner: type[object], /) -> Self:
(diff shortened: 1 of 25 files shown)


betty/property.py
```diff
@@ -20,26 +19,32 @@
     from betty.locale.localizable import ResolvableLocalizable
 
 
-class _Property[ValueGetT, ValueSetT](Attr[ValueGetT], ABC):
-    _attr_name: str
+class Property[ValueGetT, ValueSetT](Attr[ValueGetT], ABC):
+    """
+    An instance property backed by a data definition.
```
bartfeenstra, line 24: Describe how this is similar to `@property`

betty/property.py
```diff
@@ -63,18 +70,143 @@ def get(self, instance: Any, /) -> ValueGetT:
...
+    ) -> Property[ValueGetT, ValueSetT]:
+        """
+        Return a new property with the given default value factory.
+        """
+        return DefaultProperty(self, default)
+
+
+class ProxyProperty[ValueGetT, ValueSetT](Property[ValueGetT, ValueSetT]):
+    """
+    A property that proxies everything to another wrapped property.
```
bartfeenstra, line 118: s/wrapped/upstream/ here and in subclasses, for consistency with our proxy classes in other APIs.

betty/property.py
```diff
@@ -63,18 +70,143 @@ def get(self, instance: Any, /) -> ValueGetT:
...
+        # @todo
+        try:
+            return self._wrapped.get(instance)
+        except PropertyNotInitialized:
+            value = self._default(instance)
+            setattr(instance, self.name, value)
+        return value
+
+
+class AttrProperty[ValueGetT, ValueSetT](Property[ValueGetT, ValueSetT]):
```
bartfeenstra, line 207: Move this to `betty.properties`

betty/property.py
```diff
@@ -129,7 +256,7 @@ class PropertyNotInitialized(ValueError):
 
 
 @final
-class Optional[ValueGetT, ValueSetT](_Property[ValueGetT | None, ValueSetT | None]):
+class Optional[ValueGetT, ValueSetT](Property[ValueGetT | None, ValueSetT | None]):
```
bartfeenstra, line 259: Move this to `betty.properties`

PR #4194, opened 2026-05-08
Add a property (descriptor) API

bartfeenstra:
**This is being replaced by several other PRs, and is kept as a reference while work on those other PRs is being completed.**

## To do
- Consider adding a barebones variant of this API first in a separate PR. Just base class, owner and name, init. Then expand functionality in follow-ups.
- That means we can have two variants of `GetterDescriptor`: one that is a data descriptor (for attrs) and one that isn't (for service managers) without having to add subclasses specific to either of those APIs, and they'd be reusable.
- Refactor `AssetRepositoryService` into a getter? Or could the asset repository class itself not only accept paths but also plugin definitions?
- Add `AnyGettableDescriptor` that allows the return value to be changed
- Implement for `ServiceManager`
  - we currently cache internally, but that is before getters. Can we cache the outermost getter instead? We may need to subclass `GetterDescriptor` or provide a custom implementation so only the outermost will cache, and call the uncached inner getters.
  - Refactor simple service managers using getters
- Implement for `Attr`
- Rename the descriptor API to the property API for consistency with `@property`

### DIFF
--- a/betty/descriptor.py
+++ b/betty/descriptor.py
@@ -1,0 +1,310 @@
+"""
+The descriptor API.
+"""
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from inspect import getmembers
+from typing import TYPE_CHECKING, Any, Self, final, overload, override
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+
+class HasDescriptors:
+    """
+    An object that has :py:class:`descriptors <betty.descriptor.Descriptors>`.
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+        # @todo Can we do this once per class? getmembers() is quite slow and we don't want this to impact entities negatively.
+        for _, member in getmembers(type(self)):
+            # @todo This can result in descriptors to be initialized with the wrong HasDescriptions subclass?
+            if isinstance(member, Descriptor):
+                member.init_descriptor(self)
+
+
+class Descriptor[OwnerT: HasDescriptors, GetT](ABC):
+    """
+    A `descriptor <https://docs.python.org/3/howto/descriptor.html>`_.
+    """
+
+    __name: str
+    __owner: type[OwnerT]
+
+    def __set_name__(self, owner: type[OwnerT], name: str) -> None:
+        self.__owner = owner
+        self.__name = name
+
+    @final
+    @property
+    def descriptor_owner(self) -> type[OwnerT]:
+        """
+        The descriptor owner.
+        """
+        return self.__owner
+
+    @final
+    @property
+    def descriptor_name(self) -> str:
+        """
+        The descriptor name.
+        """
+        return self.__name
+
+    def init_descriptor(self, owner: OwnerT, /) -> None:  # noqa: B027
+        """
+        Initialize the descriptor.
+        """
+
+    @overload
+    def __get__(self, instance: None, owner: type[OwnerT], /) -> Self:
+        pass
+
+    @overload
+    def __get__(self, instance: OwnerT, owner: type[OwnerT] | None = None, /) -> GetT:
+        pass
+
+    @final
+    def __get__(self, instance, owner=None, /):
+        if instance is None:
+            return self
+        return self.get(instance)
+
+    @abstractmethod
+    def get(self, owner: OwnerT, /) -> GetT:
+        """
+        Get the descriptor value from the owner.
+        """
+
+    def getter(
+        self, getter: Callable[[OwnerT, GetT], GetT], /
+    ) -> Descriptor[OwnerT, GetT]:
+        """
+        Return a new descriptor with the given getter.
+        """
+        return GetterDescriptor(self, getter)
+
+    def __call__(
+        self, getter: Callable[[OwnerT, GetT], GetT], /
+    ) -> Descriptor[OwnerT, GetT]:
+        """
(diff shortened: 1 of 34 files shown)


betty/descriptor.py
```diff
@@ -0,0 +1,241 @@
...
+    """
+    A descriptor that can get a value.
+    """
+
+    @overload
+    def __get__(self, instance: None, owner: type[object], /) -> Self:
+        pass
+
+    @overload
+    def __get__(self, instance: Any, owner: type[Any] | None = None, /) -> GetT:
```
bartfeenstra: Add default implementations to the super class. In case of instance access, raise an error. `ClassOnlyDescriptor`?

bartfeenstra: We can do without the exception if we type the method to return `Self` only.

betty/descriptor.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+        # @todo Can we do this once per class? getmembers() is quite slow and we don't want this to impact entities negatively.
+        for _, member in getmembers(type(self)):
+            if isinstance(member, Descriptor):
+                member.init_descriptor(self)
+
+
+class Descriptor[InstanceT: HasDescriptors = HasDescriptors]:
```
bartfeenstra: Declare `InstanceT` in all subclasses.

betty/descriptor.py
```diff
@@ -0,0 +1,241 @@
...
+        return self.get(instance)
+
+    @abstractmethod
+    def get(self, instance: Any, /) -> GetT:
+        """
+        Get the descriptor value from the instance.
+        """
+
+    @final
+    def getter[GetterGetT](
```
bartfeenstra: Should we make these non-final to allow `Property` to allow its own subclasses that support data definitions?

betty/descriptor.py
```diff
@@ -0,0 +1,284 @@
...
+    __name: str
+    __owner: type[InstanceT]
+
+    def __set_name__(self, owner: type[InstanceT], name: str) -> None:
+        self.__owner = owner
+        self.__name = name
+
+    @final
+    @property
+    def descriptor_owner(self) -> type[InstanceT]:
```
bartfeenstra: Add a `DescriptorDefinition` dataclass to encapsulate the owner and name, so we don't have to keep prefixing methods and can have just `.descriptor` return the definition?

bartfeenstra: Can we put `ServiceProvider.id` on here as well?

betty/descriptor.py
```diff
@@ -0,0 +1,316 @@
...
+    def __delete__(self, instance: InstanceT) -> None:
+        """
+        Delete the value from the instance.
+
+        :raises AttributeError: Raised if the descriptor is not deletable.
+        """
+        self.delete(instance)
+
+
+class DescriptorProxy[InstanceT: HasDescriptors, GetT](Descriptor[InstanceT, GetT]):
```
bartfeenstra: This MUST proxy `.init_descriptor()` as well

betty/descriptor.py
```diff
@@ -0,0 +1,316 @@
...
+
+    @final
+    @property
+    def descriptor_name(self) -> str:
+        """
+        The descriptor name.
+        """
+        return self.__name
+
+    def init_descriptor(self, instance: InstanceT, /) -> None:  # noqa: B027
```
bartfeenstra: Rename to something like `.init_descriptor_owner()`, and rename "instance" to "owner" in the API, including type variables.

betty/service/__init__.py
```diff
@@ -75,63 +74,22 @@ def __init__(
...
-            return self
-
-        return self.get(instance)
-
-    def init(self, instance: ServiceProviderT, /) -> None:
-        """
-        Initialize the service.
-        """
+    @override
+    def init_descriptor(self, instance: ServiceProviderT, /) -> None:
```
bartfeenstra, line 86: Call super

betty/service/__init__.py
```diff
@@ -162,14 +118,16 @@ def _get_service_or_factory(
...
     ) -> ServiceOrFactory[ServiceProviderT, ServiceT, FactoryServiceT]:
         return getattr(
-            instance, f"_service_{self.name}_or_factory", self.__service_or_factory
+            instance,
+            f"_service_{self.descriptor_name}_or_factory",
+            self.__service_or_factory,
         )
 
     @final
     def _assert_service_not_initialized(self, instance: ServiceProviderT, /) -> None:
```
bartfeenstra, line 127: Move the assertions and the calls to the base class

betty/tests/service/test___init__.py
```diff
@@ -94,23 +61,15 @@ def my_first_service(self) -> object:
             _ServiceProvider.my_first_service.id == "_ServiceProvider.my_first_service"
         )
 
-    def test_init__initialized_already(self) -> None:
+    def test_init_descriptor__initialized_already(self) -> None:
```
bartfeenstra, line 64: Move this test to the base class

betty/tests/test_descriptor.py
```diff
@@ -0,0 +1,73 @@
+from typing import override
+
+from betty.descriptor import Descriptor, HasDescriptors
+
+
+class _DummyInstance(HasDescriptors):
```
bartfeenstra, line 6: Rename instance to owner in the tests just we need to do in the API code

PR #4204, opened 2026-05-15
Add `DefaultAttr` and `OwnerAttr.default()`

bartfeenstra:


betty/attr.py
```diff
@@ -41,9 +49,50 @@ def set(self, owner: OwnerT, value: SetT, /) -> GetT:
...
         return resolved_value
 
+    def default(self, default: Callable[[OwnerT], SetT]) -> Attr[OwnerT, GetT, SetT]:
+        """
+        Create a new attribute that proxies this one, and sets a default value.
+        """
+        return DefaultAttr(self, default)
+
+
+class DefaultAttr[OwnerT: HasProperties, GetT, SetT](Attr[OwnerT, GetT, SetT]):
```
bartfeenstra: Like with `Optional`, can we provide our own `AttrDefinition` and omit dumps if the value is equal to the default?

bartfeenstra: In fact, move `Optional`'s code here, and make optional proxy to `DefaultAttr`.

betty/project/__init__.py
```diff
@@ -792,14 +792,19 @@ class ProjectData(Data, HasProperties):
...
-        default=lambda: ProjectData._default_copyright_notice(),  # noqa: PLW0108
         resolver=CopyrightNoticeManufacturer.resolve,
     )
     """
     The project-wide copyright notice.
     """
 
+    @copyright_notice.default
+    @staticmethod
+    def copyright_notice() -> CopyrightNoticeManufacturer:  # noqa: D102
```
bartfeenstra, line 795: Manually render the docs and test that this shows the original attribute's docstring.

betty/attr.py
```diff
@@ -25,6 +24,14 @@ def __init__(self, attr: AttrDefinition[GetT], /):
         The attribute's data definition.
         """
 
+    @final
+    def _get_property_owner_attr(self, owner: OwnerT, /) -> GetT:
```
bartfeenstra: Move these helpers to a new `betty.attrs.attr.AttrAccessAttr`. But rethink the name... `Optional` and `DefaultAttr` can then extend this new class.

bartfeenstra: Ah, we can't, because nothing in `betty.attr` may depend on `bett.attrs`. should we instead move these helpers to `Attr`?

betty/attr.py
```diff
@@ -60,6 +73,45 @@ def set(self, owner: OwnerT, value: SetT, /) -> None:
...
+    """
+
+    def __init__(
+        self,
+        proxied: Attr[OwnerT, GetT, SetT],
+        default: Callable[[], SetT],
+        *args: Any,
+        **kwargs: Any,
+    ):
+        def _omit_dump(data: GetT) -> bool:
```
bartfeenstra: Add a helper to chain `omit_dump`? We are duplicating this logic here now.

betty/attr.py
```diff
@@ -60,6 +73,45 @@ def set(self, owner: OwnerT, value: SetT, /) -> None:
...
+            *args,
+            proxied=proxied,
+            **kwargs,
+        )
+        self.__default = default
+
+    @override
+    def init_property_owner(self, owner: OwnerT, /) -> None:
+        super().init_property_owner(owner)
+        self._set_property_owner_attr(owner, self.__default())
```
bartfeenstra: Only set the default if there isn't one yet. Now, what order do we do this in? Do we initialize the proxied attr before we intializer the default, or vice versa?

betty/attr.py
```diff
@@ -60,6 +73,45 @@ def set(self, owner: OwnerT, value: SetT, /) -> None:
         self.__proxied_attr.set(owner, value)
 
 
+class DefaultAttr[OwnerT: HasProperties, GetT, SetT](ProxyAttr[OwnerT, GetT, SetT]):
+    """
+    An attribute with a default value.
```
bartfeenstra: Update the docstring to clarify that the default value is stored in an owner instance attribute.

betty/attr.py
bartfeenstra: Should `AttrDefinition.omit_dump` always be a callable?

betty/attr.py
bartfeenstra: Like with `Optional`, init the proxied attribute first, and if the owner attribute still does not exist, set the default.

betty/attrs/collection_attr.py
```diff
@@ -64,4 +58,55 @@ def get(self, owner: OwnerT, /) -> MutableCollectionT:
...
+
+    @override
+    def default(
+        self, default: Callable[[], ValuesSetT] | None = None
+    ) -> Attr[OwnerT, MutableCollectionT, ValuesSetT]:
+        return _DefaultCollectionAttrAttr(self, default)
+
+
+@final
+class _DefaultCollectionAttrAttr[
```
bartfeenstra: Make this public and move it to its own module.

betty/attrs/default.py
```diff
@@ -0,0 +1,49 @@
...
+    ):
+        from betty.datas.aggregate.record import FieldDefinition
+
+        super().__init__(
+            proxied,
+            attr=FieldDefinition(
+                proxied.field.data,
+                label=proxied.field.label,
+                description=proxied.field.description,
+                omit_load=proxied.field.omit_load,
```
bartfeenstra: This must be `True`

betty/attrs/default_collection.py
```diff
@@ -0,0 +1,57 @@
...
+        default: Callable[[], ValuesSetT],
+        /,
+    ):
+        super().__init__(
+            proxied,
+            attr=FieldDefinition(
+                proxied.field.data,
+                label=proxied.field.label,
+                description=proxied.field.description,
+                omit_load=proxied.field.omit_load,
```
bartfeenstra: This must be True

PR #4209, opened 2026-05-16
Add `ProxyProperty` and `ProxyAttr`

bartfeenstra:


betty/attrs/optional.py
```diff
@@ -45,6 +45,12 @@ def __set_name__(self, owner: type[OwnerT], name: str) -> None:
 
     @override
     def get(self, owner: OwnerT, /) -> GetT | None:
+        # @todo Refactor this.
+        # @todo Get rid of the exception altogether.
+        # @todo Use init_property_owner() to set the default.
+        # @todo
+        # @todo Maybe do this in YAPR (Yet Another PR)?
```
bartfeenstra: Fix this first

betty/property.py
```diff
@@ -92,3 +92,16 @@ def get(self, owner: OwnerT, /) -> GetT:
         """
         Get the property value from the owner.
         """
+
+
+class ProxyProperty[OwnerT: HasProperties, GetT](Property[OwnerT, GetT]):
+    def __init__(self, *args, **kwargs):
```
bartfeenstra: Allow access to the proxied property, but only privately. Subclasses can keep their own private, typed references. Using multiple references, each with their own type, means we don't have to be generic over the proxied type, which could become messy.

betty/property.py
```diff
@@ -92,3 +92,16 @@ def get(self, owner: OwnerT, /) -> GetT:
         """
         Get the property value from the owner.
         """
+
+
+class ProxyProperty[OwnerT: HasProperties, GetT](Property[OwnerT, GetT]):
```
bartfeenstra, line 98: Also proxy `.get()`

betty/attrs/optional.py
bartfeenstra: Blocked by https://github.com/bartfeenstra/betty/pull/4210

PR #4210, opened 2026-05-16
Remove `AttrNotInitialized`

bartfeenstra:


betty/attrs/attr.py
```diff
@@ -54,16 +52,4 @@ def __init__(
     @final
     @override
     def get(self, owner: OwnerT, /) -> GetT:
-        value = cast(
-            GetT | Void,
-            getattr(owner, f"_{self.property.name}", Void),
-        )
-        if value is Void:
-            if self._default is None:
```
bartfeenstra, line 62: Add this back but just catch `AttributeError` for now

PR #4222, opened 2026-05-17
Add `ResolvableDataDefinition`

bartfeenstra:
Scale/down reintroduce the changes if we cannot locate the root cause of the type failures.

### DIFF
--- a/betty/datas/aggregate/collection/__init__.py
+++ b/betty/datas/aggregate/collection/__init__.py
@@ -8,7 +8,7 @@ from abc import ABC, abstractmethod
 from collections.abc import Callable, Collection, Iterable
 from typing import TYPE_CHECKING, Any, final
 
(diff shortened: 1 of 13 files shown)


betty/datas/aggregate/collection/__init__.py
```diff
@@ -8,12 +8,12 @@
...
 from typing import TYPE_CHECKING, Any, final
 
-from betty.data import DataDefinition
+from betty.data import DataDefinition, ResolvableDataDefinition, resolve_data_definition
 from betty.datas.aggregate import AggregateDefinition
 from betty.indicator.selector import Element
 
 if TYPE_CHECKING:
-    from betty.data import Data
+    from betty.data import Data as Data
```
bartfeenstra, line 16: Remove and check for other such errors

PR #4265, opened 2026-05-20
Refactor entity associations onto `betty.attr`

### LINKED ISSUE #2778: To-many entity resolvers do not persist order

### LINKED ISSUE #3533: Merge entity references and entity resolvers

### LINKED ISSUE #3581: Refactor entity associations onto `betty.property`
## To do
- `LinkedDataDumpableWithSchemaJsonLdObject.linked_data_schema` was updated to always make properties optional as an intermediate solution until this issue is fixed. Ensure properties are set to required depending on their aggregate element definitions.
- remove `ToZero` associations in favor of the new `Optional` descriptor.

## Blocked by
- https://github.com/bartfeenstra/betty/issues/3580
- https://github.com/bartfeenstra/betty/issues/3609

### LINKED ISSUE #3755: Refactor entity associations onto `betty.collections`
## To do
- type on `KeyedCollection`
- we need event handling support, specifically to act on individual associates being added or removed, to set bidirectionality. We **could** (maybe not should?) use value resolvers to add new associates. Also we could use the main collection resolver but we'd have to do all the comparisons ourselves. 

## Blocked by
- https://github.com/bartfeenstra/betty/issues/3730

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3581, and https://github.com/bartfeenstra/betty/issues/3755, and https://github.com/bartfeenstra/betty/issues/2778


betty/entity/association.py
```diff
@@ -467,296 +368,189 @@ def __init__(
...
-    @property
-    def associate_attr_name(self) -> str:
-        """
-        The association's attribute name on the associate type.
-        """
-        return self._associate_attr_name
+        return hash((super().__hash__(), self.associate_attr_name))
 
-    def inverse(self) -> _BidirectionalAssociation[AssociateT, OwnerT]:
+    def inverse(
```
bartfeenstra: Cache this

bartfeenstra: `@cached_property`?

bartfeenstra: No! Because that allows setting values and we explicitly do not want that here.

betty/entity/association.py
```diff
@@ -207,229 +180,195 @@ def disassociate(self, owner: OwnerT, associate: AssociateT, /) -> None:
...
-
-    def __get__(self, instance: OwnerT | None, owner: type[OwnerT] | None = None, /):
-        if instance is None:
-            return self
+    """
+    A *-to-many entity association.
+    """
+
+    @override
+    def get(self, owner: _OwnerT, /) -> EntityCollection[_AssociateT]:
```
bartfeenstra: Refactor this to use existing attribute functionality. Maybe subclass `ProxyProp` and proxy `CollectionOwnerAttr`?

betty/datas/optional.py
```diff
@@ -18,11 +19,15 @@ class OptionalDefinition[DataClsT](DataDefinition[DataClsT | None]):
     """
 
     def __init__(self, wrapped: DataDefinition[DataClsT], /):
+        try:
```
bartfeenstra: We should revert this as the assumption in the data API is that everything is portable, hence the exception and no `None` return value. We should instead use temporary dummy porters for our associations.

bartfeenstra: Is this where entity references come on stage again? Should we move them back into the entity API?

bartfeenstra: I mean, ideally we add full serialization support now straight away. Probably nota bad idea because there are some issues we haven't thought out yet, e.g. how to resolve/hydrate loaded associations.

betty/entity/association.py
```diff
@@ -39,724 +42,115 @@ async def _generate_associate_url(project: Project, associate: Entity, /) -> str
...
+    | Callable[[Association[OwnerT, AssociateT], OwnerT], AssociateT]
+    | Callable[[Project, Association, OwnerT], AssociateT]
+)
+type Associate[OwnerT: Entity, AssociateT: Entity] = (
+    AssociateT | AssociateResolver[OwnerT, AssociateT]
+)
 
 
-def resolve(*entities: Entity) -> None:
+def resolve(project: Project, *entities: Entity) -> None:
```
bartfeenstra: We need this signature so things like loaders can resolve all their entities in one go.

We also need a version to just resolve a single resolver, with the logic for determining which arguments to pass on.

bartfeenstra: We also need a variant of that second signature for the main entity API.

bartfeenstra: Code elsewhere that types on `EntityReference` probably needs refactoring to accept `type ResolvableEntity[EntityT: Entity] = EntityT | EntityResolver[EntityT]` instead.

betty/entity/associations/to_many.py
```diff
@@ -0,0 +1,245 @@
...
+
+if TYPE_CHECKING:
+    from betty.collection.keyed import MutableKeyedCollection
+    from betty.locale.localizable import ResolvableLocalizable
+    from betty.portable import PortableData
+    from betty.project import Project
+
+
+@final
+class ToManyAssociates[OwnerT: Entity, AssociateT: Entity](
```
bartfeenstra: Rename to `ToManyCollection`.

betty/associations/to_many.py
```diff
@@ -0,0 +1,245 @@
...
+if TYPE_CHECKING:
+    from betty.collection.keyed import MutableKeyedCollection
+    from betty.locale.localizable import ResolvableLocalizable
+    from betty.portable import PortableData
+    from betty.project import Project
+
+
+@final
+class ToManyAssociates[OwnerT: Entity, AssociateT: Entity](
+    MutableCollection[AssociateT],
```
bartfeenstra: Make this a sequence too? Just a regular one, not mutable.

betty/associations/to_many.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def add(self, *associates: Associate[OwnerT, AssociateT]) -> None:
+        """
+        Add the given associate.
+        """
+        for associate in associates:
+            is_entity = isinstance(associate, Entity)
+            if not is_entity:
+                self._resolved = False
+            elif associate in tuple(
```
bartfeenstra: Move this check outside the loop.

betty/entity/associations/to_many.py
```diff
@@ -0,0 +1,245 @@
...
+        /,
+    ):
+        self._association = association
+        self._owner = owner
+        self._resolved = True
+        self._associates = []
+
+    def add(self, *associates: Associate[OwnerT, AssociateT]) -> None:
+        """
+        Add the given associate.
```
bartfeenstra: Plural

betty/entity/associations/to_one.py
```diff
@@ -0,0 +1,252 @@
...
+class ToOne[OwnerT: Entity, AssociateT: Entity](
+    Association[OwnerT, AssociateT, AssociateT, ToOneAssociate[OwnerT, AssociateT]]
+):
+    """
+    A *-to-one entity association.
+    """
+
+    def __init__(
+        self,
+        associate_type_name: str,
```
bartfeenstra: Allow types as well as names.

betty/associations/to_many.py
```diff
@@ -0,0 +1,245 @@
...
+        Iterable[Associate[OwnerT, AssociateT]],
+    ]
+):
+    """
+    A *-to-many entity association.
+    """
+
+    def __init__(
+        self,
+        associate_type_name: str,
```
bartfeenstra: Allow types as well as names.

betty/entities/citation.py
```diff
@@ -77,12 +76,12 @@ class Citation(HasAnyDate, HasFileReferences, HasPrivacy, HasLinks):
...
         *,
-        source: ToOneAssociate[Source],
+        source: Associate[Self, Source],
         id: str | None = None,  # noqa: A002
-        facts: ToManyAssociates[HasCitations] = (),
+        facts: Iterable[Associate[Self, HasCitations]] = (),
         location: ResolvableLocalizable | None = None,
         date: AnyDate | None = None,
-        file_references: ToManyAssociates[FileReference] = (),
+        file_references: Iterable[Associate[Self, FileReference]] = (),
```
bartfeenstra: Add `ToManyAssociates` again

betty/entities/event.py
```diff
@@ -99,13 +96,13 @@ def __init__(
         id: str | None = None,  # noqa: A002
         event_type: EventType | None = None,
         date: AnyDate | None = None,
-        file_references: ToManyAssociates[FileReference] = (),
```
bartfeenstra: In a separate PR, rename `HasFileReferences.file_references` to `.files`.

betty/associations/to_many.py
bartfeenstra: All we'd need to implement are `__reversed__()` and `__getitem__()` anyway, which are easy and sensible.

betty/entity/associations/to_one.py
```diff
@@ -0,0 +1,259 @@
...
+    async def dump_linked_data_for(
+        self, project: Project, target: OwnerT, /
+    ) -> PortableData:
+        if getattr(target, self.prop.owner_attr):
+            return None
+        return await super().dump_linked_data_for(project, target)
+
+
+@final
+class _ToOnePorter[AssociateT: Entity](Porter):
```
bartfeenstra: Rename this to `AssociatePorter` and move it into the main API.

betty/entity/associations/to_one.py
```diff
@@ -0,0 +1,259 @@
...
+        associate_type_name: str,
+        *,
+        bi: str | None = None,
+        description: ResolvableLocalizable | None = None,
+        label: ResolvableLocalizable,
+    ):
+        super().__init__(
+            FieldDefinition(
+                DataDefinition(
+                    Entity, description=description, label=label, porter=_ToOnePorter()
```
bartfeenstra: In a separate PR, can we make entities defined data (just with empty definitions for now)? Then in this PR we can at least finish the associations' data definitions.

betty/gramps.py
```diff
@@ -191,27 +189,15 @@ def __str__(self) -> str:
         return f"{self.entity_type.value} ({self.entity_id})"
 
 
-class _ToOneResolver[EntityT: Entity](ToOneResolver[EntityT]):
+class _Resolver[EntityT: Entity]:
```
bartfeenstra: Remove the resolver classes and replace their usage with `functools.partial(handles_to_entities.get, "my-first-gramps-handle")`

betty/associations/to_many.py
```diff
@@ -0,0 +1,121 @@
...
+        *,
+        associate_attr: str | None = None,
```
bartfeenstra, line 47: Make the attr name positional for all associations.

bin/test.py
```diff
@@ -10,7 +10,10 @@
 check_call([sys.executable, path.join("bin", "clean-build.py")])
 if "BETTY_TEST_SKIP_RUFF" not in environ or not environ["BETTY_TEST_SKIP_RUFF"]:
     check_call([sys.executable, path.join("bin", "test-ruff.py")])
-check_call([sys.executable, path.join("bin", "test-ty.py")])
+# ty is experiencing several stack overflow issues. At the time of writing, on our GitHub Actions, these issues only
```
bartfeenstra: Create a follow-up issue if this ends up being merged.

betty/collections/to_many.py
```diff
@@ -0,0 +1,140 @@
...
+            if not is_entity:
+                self._resolved = False
+            elif associate in tuple(
+                existing_associate
+                for existing_associate in self
+                if isinstance(existing_associate, Entity)
+            ):
+                continue
+            self._associates.append(associate)
+            if is_entity and (associate_attr := self._association.associate_attr):
```
bartfeenstra: If we repeat the instance check here, we may be able to get rid of the ty ignore right below

betty/collections/to_many.py
```diff
@@ -0,0 +1,140 @@
...
+                )
+
+    def remove(self, *associates: AssociateT) -> None:
+        """
+        Remove the given associates.
+        """
+        # @todo Raise a helpful error.
+        assert self._resolved
+        if associates:
+            bi = self._association.associate_attr
```
bartfeenstra: Rename these occurrences

betty/tests/collections/test_to_many.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+
+
+class _BiAssociate(Entity):
+    owners = ToMany(
+        "betty.tests.collections.test_to_many._Owner", "associates", label="-"
+    )
+
+
+class TestToManyCollection:
```
bartfeenstra, line 55: Check coverage of all execution paths!

betty/entity/__init__.py
```diff
@@ -137,4 +160,63 @@ def __init__(
...
+
+
+@final
+@ObjectDefinition(
+    label=_("Entity reference"),
+    samples=[
+        lambda: Sample(EntityReference("person", "123"), label="Default"),
+    ],
+)
+class EntityReference[EntityT: Entity = Entity](Data, HasProps):
```
bartfeenstra: Check where we import this, but see if we can move this back to `betty.datas`

betty/association.py
```diff
@@ -0,0 +1,203 @@
...
+            if isinstance(associate_attr, Association)
+            else associate_attr
+        )
+        self.__associate_attr: (
+            Association[AssociateT, OwnerT, Any, Any] | None | VoidType
+        ) = associate_attr if isinstance(associate_attr, Association) else Void
+
+    @final
+    @property
+    def associate_attr(
```
bartfeenstra, line 67: If no associate attr name was declared, check if the associate type has an association that points to the owner type and has an associate attr name pointing to this owner association.

PR #4266, opened 2026-05-24
Property registration

bartfeenstra:


betty/attr.py
```diff
@@ -4,15 +4,26 @@
...
 
 
-class Attr[OwnerT: HasProperties, GetT, SetT](SettableProperty[OwnerT, GetT, SetT]):
+class HasAttrs(HasProperties):
+    """
+    An object that has :py:class:`attributes <betty.attr.Attr>`.
+    """
+
+    _attrs: ClassVar[set[Attr[HasAttrs, Any, Any]]] = set()
+    attrs: ClassVar[Set[Attr[HasAttrs, Any, Any]]] = _attrs
```
bartfeenstra: Do we really need these? Can we use `[property for property in HasProperties.properties if isinstance(property, Attr)]`?

bartfeenstra: Or make `.properties` `.properties(cls: type[Property] = Property)` so it works for any and all sub-APIs.

bartfeenstra: Except... Test that return typing works for us. Because `.properties(Attr)` must return `Intersection[Property[Self, Any, Any], Attr]`

betty/property.py
```diff
@@ -6,33 +6,27 @@
...
-    from collections.abc import Sequence
+    from collections.abc import Set
 
 
 class HasProperties:
     """
     An object that has :py:class:`properties <betty.property.Property>`.
     """
 
+    _properties: ClassVar[set[Property[Any, Any]]] = set()
```
bartfeenstra: `Property[Self, ...]`

betty/attr.py
```diff
@@ -6,13 +6,29 @@
...
+    ProxyProperty,
+    SettableProperty,
+)
 
 if TYPE_CHECKING:
     from betty.datas.aggregate.record import FieldDefinition
 
 
-class Attr[OwnerT: HasProperties, GetT, SetT](SettableProperty[OwnerT, GetT, SetT]):
+# @todo Do we need this still?
```
bartfeenstra: Do we?

betty/property.py
```diff
@@ -97,6 +90,36 @@ def get(self, owner: OwnerT, /) -> GetT:
         """
 
 
+@final
+class PropertyRegistry[OwnerT: HasProperties, PropertyT: Property]:
+    """
+    A registry of all properties on classes.
```
bartfeenstra: Remove "all"

betty/property.py
```diff
@@ -97,6 +90,36 @@ def get(self, owner: OwnerT, /) -> GetT:
...
+    def register(
+        self,
+        owner: type[OwnerT],
+        property: PropertyT,  # noqa: A002
+        /,
+    ) -> None:
+        """
+        Register a new property for an owner.
+        """
+        self._registry[owner].append(property)
```
bartfeenstra: Error if a property with that name was registered already.

bartfeenstra: And test

betty/property.py
```diff
@@ -97,6 +90,37 @@ def get(self, owner: OwnerT, /) -> GetT:
...
+    """
+
+    def __init__(self):
+        self._registry = defaultdict(list)
+
+    def get(self, owner: type[OwnerT] | OwnerT, /) -> Iterable[PropertyT]:
+        """
+        Get the properties for an owner.
+        """
+        for mro in (owner if isinstance(owner, type) else type(owner)).__mro__:
```
bartfeenstra: Instead:
```python
yield from self._registry[owner]
for base in owner.__bases__:
  yield from self.get(base)
```

betty/property.py
```diff
@@ -97,6 +90,26 @@ def get(self, owner: OwnerT, /) -> GetT:
...
+    """
+    Get all properties on an object.
+    """
+    return _get_properties(owner if isinstance(owner, type) else type(owner))
+
+
+@cache
+def _get_properties[OwnerT: HasProperties](
+    owner: type[OwnerT], /
+) -> Iterable[Property[OwnerT, Any]]:
```
bartfeenstra: Return a sequence as this is the internal, cached function

betty/property.py
```diff
@@ -97,6 +90,26 @@ def get(self, owner: OwnerT, /) -> GetT:
         """
 
 
+def get_properties[OwnerT: HasProperties](
```
bartfeenstra: Move this onto `HasProperties`

betty/attr.py
```diff
@@ -24,7 +42,18 @@ def __init__(self, field: FieldDefinition[OwnerT, GetT], /):
         """
 
 
-class ProxyAttr[OwnerT: HasProperties, GetT, SetT](
+def get_attrs[OwnerT: HasAttrs](
```
bartfeenstra: Move this onto `HasAttrs`

bartfeenstra: Rename to `attrs()` and make it a class method

betty/property.py
bartfeenstra: Rename to `properties()` and make it a class method

betty/attr.py
bartfeenstra: Update the docstring

betty/property.py
bartfeenstra: Update the docstring

betty/property.py
```diff
@@ -97,6 +90,26 @@ def get(self, owner: OwnerT, /) -> GetT:
...
+    owner: type[OwnerT] | OwnerT, /
+) -> Iterable[Property[OwnerT, Any]]:
+    """
+    Get all properties on an object.
+    """
+    return _get_properties(owner if isinstance(owner, type) else type(owner))
+
+
+@cache
+def _get_properties[OwnerT: HasProperties](
```
bartfeenstra: Make this a class method

PR #4277, opened 2026-06-02
change aioresponses to aiointercept

Polandia94:
Solve: https://github.com/bartfeenstra/betty/pull/4276

Changed the test library from aioresponses to aiointercept. Aiointercept is a library that of wich I'm the maintainer. It has a API similar to aioresponses but it create a real server and does real request, against localhost


bartfeenstra:
Please don't spam.

PR #4432, opened 2026-06-25
Reduce the JSON schema API

### LINKED ISSUE #3829: Remove the JSON schema API

### LINKED ISSUE #4164: Add `LinkedDataDefinition`
Refactor the linked data API to be similar to defined data.

## June 21 thoughts
- Add `DataDefinition.linked_data_porter` and make it work similar to `DataDefinition.porter`. Keeping it inside data definitions keeps the change simple for now while solving the issues we originally set out to solve.
- Can we set default linked data porters that just proxy regular porters, except they return `Void` if the data is not public? How to handle privacy for attributes? Some attributes are always public, like associations. Others depend on the owner. Should we add this to `RecordDefinition`, an iterable of attribute names that are public (but this can never override any explicit/determined privacy declared by individual attributes).
- Should attribute privacy be a thing? Or even `FieldDefinition.privacy`?
- Also, privacy in which context? When performing which action? An entity association may be private if the check is "can we follow this association to publish an associate's attribute", but public when dumping linked data (because that only includes anonymous data such as URLs). In many other cases an attribute's privacy may be "undetermined" and it will dump linked data but omit sensitive fields (e.g. what we do when dumping entities to linked data).

## To do
- Add `LinkedDataDefinition(DataDefinition)`.
- Linked data definitions should have porters like regular data definitions, but for now we only specify `.dump() -> PortableData | VoidType` on them.
- Add `ProxyDataDefinition`?
- Add a porter/dumper implementation that dispatches to `LinkedDataDumpable`, similar to how we dispatch to `Portable`
- Rename `LinkedDataDumpable` to `LinkedDataPortable`
- Do we want a temporary BC later to `LinkedDataDumper`?
- Add `ProxyLinkedDataDefinition(LinkedDataDefinition)` which takes `data: DataDefinition` as well as a linked data porters/dumper. That way we can easily add linked data functionality to existing data definitions.
- Wherever entities currently dispatch linked data dumps or schemas to other code (such as mixins, static translations, etc.), refactor this to be provided by `Attr` class members.
- implement for `Optional`
- Remove `LinkedDataDumper` (this is replaced by linked data porters)

## Blocked by 
- https://github.com/bartfeenstra/betty/issues/3829
- https://github.com/bartfeenstra/betty/issues/4168
- https://github.com/bartfeenstra/betty/issues/4169
- https://github.com/bartfeenstra/betty/issues/3581

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/3829 and https://github.com/bartfeenstra/betty/issues/4164

## To do
- Finish `RecordLinkedDataPorter`. Ensure it takes owner as well as field privacy into account.
- Make all schemas for linked data private and move them into wherever that linked data is dumped. 
- Consider introducing `DataDefinition.linked_data_porter` in this PR. Without it, we would need to create new `Attr` subclasses for each attribute that requires linked data porting, where we don't want to do this on the owner (looking at you, `new_localizable_attr()`)
- ATTRS WHO DO NOT HANDLE THEIR OWN LINKED DATA YET
  - plugin instances
    - place type
    - person gender
    - file copyright notice
    - file license
    - event type
  - place coordinates
  - plain strings with contexts (person name individual and affiliation, link rel)
  - virtual properties (make these all virtual attrs in a separate PR first?)
    - event.eventAttendanceMode
    - event.eventStatus
    - date.iso8601 (rename to date.date?)


betty/json_schema.py
```diff
@@ -4,19 +4,18 @@
...
-
 if TYPE_CHECKING:
-    from betty.localizable import ResolvableLocalizable
+    from collections.abc import Mapping
+
+    from betty.portable import PortableData, PortableMapping
 
 
+@final
 class Schema:
```
bartfeenstra: Make this a dataclass

betty/linked_data.py
```diff
@@ -5,165 +5,91 @@
...
-
-        return portable
-
-
-class LinkedDataDumper[
-    T,
-    SchemaT: Schema = Schema,
-    PortableDataT: PortableData = PortableData,
-](ABC):
+class LinkedDataPorter[T](ABC):
```
bartfeenstra, line 53: Make sure that attributes no longer subclass this. It's for porters only.

data/asset/builtin/templates/linked-data.html.j2
```diff
@@ -1,8 +1,8 @@
 {% set linked_data=[
     document.breadcrumbs.dump_linked_data(project)
 ] %}
-{% if document.resource is linked_data_dumpable %}
-    {% do linked_data.append(document.resource.dump_linked_data(project)) %}
+{% if document.resource is entity_plugin %}
+    {% do linked_data.append(document.resource.data().linked_data_porter.dump(project)) %}
```
bartfeenstra, line 5: This is tricky because we can't handle exceptions here and it's not guaranteed that linked data porters are available

bartfeenstra: Should we make the porters on data definitions optional then?

PR #4442, opened 2026-06-27
Refactor the date API

### LINKED ISSUE #4451: Multi-calendar support
## Definition of Done
- Use ISO 8601 by default 
- Support all calendars Gramps supports. 
- Support conversions between any calendar and ISO 8601
- Date validation per calendar

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/4451

## To do
- Add `DateExpressionFormatter` that is completely agnostic of `betty.date` Can this replace `FormatDatetimeDatetime` as well? Ideally we'd have (overloaded) methods for non-`Localizable` types.
- Add `Localizer.date: DateExpressionFormatter`
- Calculating the earliest and latest possible variants of an incomplete date requires knowledge of which date the given new year started in. For some calendars we can set a reasonable default. For others, this is a specific bit of 'locale' information that goes with the date, because historically, whatever was the New Year's day in a given place in a given year changed a lot. Gramps also addresses this by allowing dates to specify their own New Year's day (we would need a month and a day: `Date.new_years_day: tuple[int, int] | None`).
- Can we use this, and a new `AnyDateSeries` (a series of dates, one or more of which may be applicable), to encode dual-dated dates when parsing them from Gramps?
- When compiling comparison date parts, do so in ISO 8601 **as well as** the date's original calendar, so the date can at least be compared to other dates on the same calendar.
- What to do with dates on different calendars that cannot be compared? We don't want to exclude them anywhere.
- Always include dates with missing parts when sorting. Just decide whether to put them at the start, or at the end.
- In `Date._get_comparable_start()` we assume that the start of the new year is on the first day of the first month, but that is absolutely not a given thing. We should delegate this to `Calendar` implementations, but bar the ISO 8601 ones, we may not be able to fill in missing months, only missing days.
- Define the `date` linked data field in terms of ISO 8601 but with the caveat that we allow four OR MORE year digits and that years are ALWAYS prefixed with a plus or minus sign
- Consider adding support for multiple date formatting variants: short, medium, long (with day of the week)?
- Ensure Gramps can load dates on any of the other calendars we now support. Also support loading dual-dated dates!
- Speaking of, which other calendars to add besides Julian? Gramps supports Gregorian, Hebrew, French Republican, Julian, Islamic, Persian, and Swedish. Just all of the calendars supported by convertdate?
- Fix `@todo`s

### DIFF
--- a/betty/calendar.py
+++ b/betty/calendar.py
@@ -1,0 +1,138 @@
+"""
+The calendar API.
+"""
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING, Any, final
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    from betty.localizable import Localizable
+    from betty.machine_name import MachineName
+
+
+class Calendar(ABC):
+    """
+    A calendar.
+    """
+
+    @final
+    def __init_subclass__(cls, **kwargs: Any):
+        assert cls.__module__.startswith("betty.calendars."), (
+            "Third-party calendars are not supported."
+        )
+        super().__init_subclass__(**kwargs)
+
+    @classmethod
+    @abstractmethod
+    def id(cls) -> MachineName:
+        """
+        The unique calendar ID.
+        """
+
+    @classmethod
+    @abstractmethod
+    def label(cls) -> Localizable:
+        """
+        The human-readable calendar label.
+        """
+
+    @classmethod
+    @abstractmethod
+    def public_label(cls) -> Localizable:
+        """
+        The human-readable public calendar label.
+        """
+
+    @classmethod
+    @abstractmethod
+    def years(cls) -> Sequence[int]:
+        """
+        Get the years on this calendar.
+        """
+
+    @classmethod
+    @abstractmethod
+    def months(cls) -> int:
+        """
+        Get the number of months in a year.
+        """
+
+    @classmethod
+    @abstractmethod
+    def days(cls, year: int | None, month: int | None, /) -> int:
+        """
+        Get the number of days for the given month, or the maximum number of days any month could have.
+        """
+
+    # @todo Should we be using this at all? It's too simplistic, because it cannot handle the new year not
+    # @todo being on the first day of the frst month.
+    # @todo
+    # @todo Can we refactor this using cls.earliest() and cls.latest()?
+    # @todo
+    @classmethod
+    def validate(
+        cls, year: int | None, month: int | None, day: int | None, /
+    ) -> bool | None:
+        """
+        Validate the given date against this calendar.
+
+        Return ``True`` if all date parts are present and valid. Return ``False`` if any date part is invalid. Return
+        ``None`` otherwise, indicating that any given date was valid, but not all data was given.
+        """
+        if year is not None and year not in cls.years():
+            return False
+        if month is not None:
+            if month < 1:
+                return False
+            if month > cls.months():
+                return False
+        if day is not None:
+            if day < 1:
+                return False
+            if day > cls.days(year, month):
+                return False
+        if year is None:
+            return None
+        if month is None:
+            return None
+        if day is None:
+            return None
+        return True
+
+    @classmethod
+    @abstractmethod
+    def to_proleptic_iso8601(
+        cls, year: int, month: int, day: int, /
+    ) -> tuple[int, int, int]:
+        """
+        Convert a date on this calendar to one on the :py:class:`proleptic ISO 8601 calendar <betty.calendars.iso8601.ProlepticIso8601>`.
+        """
+
+    @classmethod
+    def earliest(
+        cls, year: int | None, month: int | None, day: int | None, /
+    ) -> tuple[int | None, int | None, int | None]:
+        """
+        Try to fill in the missing date parts with the earliest possible values.
+        """
+        return year, month, day
+
+    @classmethod
+    def latest(
+        cls, year: int | None, month: int | None, day: int | None, /
+    ) -> tuple[int | None, int | None, int | None]:
+        """
+        Try to fill in the missing date parts with the latest possible values.
+        """
+        return year, month, day
+
+
+def to_year_zero(year: int, /) -> int:
+    """
+    Convert a year from a numbering without year zero to a numbering with year zero.
+    """
+    raise NotImplementedError
--- a/betty/date.py
+++ b/betty/date.py
@@ -1,459 +1,493 @@
 """
-Localize dates.
+The date API.
 """
 
 from __future__ import annotations
 
-import calendar
 import datetime
-import operator
-from contextlib import suppress
-from functools import total_ordering
-from typing import TYPE_CHECKING, Any, final, override
+from abc import ABC, abstractmethod
+from operator import not_
+from typing import TYPE_CHECKING, Any, Final, Self, TypeGuard, final, overload, override
 
 from babel import dates
 
-from betty.attrs.owner import OwnerAttr
-from betty.data import Data
+from betty import calendars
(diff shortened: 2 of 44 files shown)


betty/date.py
```diff
@@ -38,111 +38,92 @@ class IncompleteDateError(ValueError):
...
-    }
+    __slots__ = ("year", "month", "day", "fuzzy", "normalized")
 
-    year = OwnerAttr(IntDefinition(label=_("Year"))).optional
-    month = OwnerAttr(IntDefinition(label=_("Month"))).optional
-    day = OwnerAttr(IntDefinition(label=_("Day"))).optional
-    fuzzy = OwnerAttr(BoolDefinition(label=_("Fuzzy"))).default(lambda: False)
+    _unknown_localize_format = _("unknown date")
+    _localize_formats: Final[dict[tuple[bool, bool, bool], Localizable]] = {
+        (True, True, True): _("MMMM d, y"),
```
bartfeenstra: Move these patterns to `Localizer`? Also make sure any date is always localizable.

betty/date.py
```diff
@@ -7,142 +7,119 @@
...
-    def parts(self) -> tuple[int | None, int | None, int | None]:
+        self.fuzzy: Final[bool] = fuzzy
         """
-        The date parts: a 3-tuple of the year, month, and day.
+        Whether the year, month, and/or day are fuzzy, e.g. not exactly known. 
         """
-        return self.year, self.month, self.day
 
-    def to_range(self) -> DateRange:
+        self.normalized: Final[AnyDate | None] = self._normalize()
```
bartfeenstra: Make private and rename to `self._compiled`.

betty/date.py
```diff
@@ -186,178 +193,143 @@ def __le__(self, other: Any) -> bool:
...
         super().__init__()
-        self.start = start
-        self.start_is_boundary = start_is_boundary
-        self.end = end
-        self.end_is_boundary = end_is_boundary
+        self.start: Final[Date | None] = start
+        self.start_is_boundary: Final[bool] = start_is_boundary
+        self.end: Final[Date | None] = end
+        self.end_is_boundary: Final[bool] = end_is_boundary
+        self.normalized: Final[DateRange | None] = self._normalize()
```
bartfeenstra: Make private and rename to `self._compiled`.

betty/date.py
```diff
@@ -4,448 +4,388 @@
...
+            localized = _("around {date}").format(date=localized).localize(localizer)
+        if not isinstance(self.calendar, Gregorian):
+            localized = (
+                _("{date} ({calendar})")
+                .format(date=localized, calendar=self.calendar.label)
+                .localize(localizer)
+            )
+        return localized
+
+    def __truediv__(self, other: Any) -> AnyDate:
```
bartfeenstra: Add a private base class for these shared implementations

betty/date.py
```diff
@@ -1,459 +1,514 @@
...
 
+type _GregorianDateParts = tuple[int | None, int | None, int | None, bool]
 
-@final
-class IncompleteDateError(ValueError):
-    """
-    Raised when a date-like was unexpectedly incomplete.
-    """
 
+class _AnyDate(Localizable, Data[ObjectDefinition]):
```
bartfeenstra: Make this public and prevent third-party subclasses.

betty/date.py
```diff
@@ -1,459 +1,514 @@
...
-            other.start,
-            other.end,
-            other.start_is_boundary,
-            other.end_is_boundary,
-        )
+    def localize(self, localizer: Localizer, /) -> LocalizedStr:
+        return AllEnumeration(*self.dates).localize(localizer)
 
 
+# @todo Purposefully omit DateSeries for now as we develop it and see how useful it becomes.
```
bartfeenstra: Remove this

betty/date.py
```diff
@@ -1,459 +1,514 @@
...
-                if another <= self.end:
-                    return True
-        return False
-
-    def _get_comparable_date(self, date: Date | None, /) -> Date | None:
-        if date and date.comparable:
-            return date
-        return None
+@final
+class DateSeries(_AnyDate):
```
bartfeenstra: Add a data definition

betty/date.py
```diff
@@ -1,459 +1,514 @@
...
+            BoolDefinition(label=_("Fuzzy")),
+            omit_load=True,
+            omit_dump=lambda data: data is False,
+        ),
+    },
+    factory=lambda *, year, month, day, fuzzy: Date(year, month, day, fuzzy=fuzzy),
     samples=[
-        lambda: Sample(Date(), label="Minimal", size=Size.MINIMAL),
+        lambda: Sample(Date(1), label="Minimal", size=Size.MINIMAL),
         lambda: Sample(Date(1970, 1, 1, fuzzy=True), label="Full", size=Size.FULL),
```
bartfeenstra: Add a calendar sample, and to full

betty/date.py
```diff
@@ -1,459 +1,682 @@
...
+from typing import TYPE_CHECKING, Any, Final, Self, final, overload, override
 
 from babel import dates
 
-from betty.attrs.owner import OwnerAttr
-from betty.data import Data
+from betty import calendars
+from betty.assertions.if_else import assert_if_else
+from betty.calendars import default_calendar
+from betty.calendars.gregorian import Gregorian
```
bartfeenstra: Use `betty.calendara.default_calendar` instead.

betty/calendar.py
```diff
@@ -0,0 +1,122 @@
...
+    @classmethod
+    @abstractmethod
+    def years(cls) -> Sequence[int]:
+        """
+        Get the years on this calendar.
+        """
+
+    @classmethod
+    @abstractmethod
+    def months(cls, year: int | None, /) -> int:
```
bartfeenstra: Remove the year argument. All our calendars use a fixed number of months.

betty/calendar.py
```diff
@@ -0,0 +1,134 @@
...
+    @classmethod
+    @abstractmethod
+    def public_label(cls) -> Localizable:
+        """
+        The human-readable public calendar label.
+        """
+
+    @classmethod
+    @abstractmethod
+    def years(cls) -> Sequence[int]:
```
bartfeenstra, line 52: Can we replace this with only an 'earliest (possible) year'?

betty/date.py
```diff
@@ -1,459 +1,491 @@
...
 
-_LOCALIZE_DATE_PART_FORMATS: Mapping[tuple[bool, bool, bool], Localizable] = {
-    (True, True, True): _("MMMM d, y"),
-    (True, True, False): _("MMMM, y"),
-    (True, False, False): _("y"),
-    (False, True, True): _("MMMM d"),
-    (False, True, False): _("MMMM"),
-}
 
+class DateExpression(Localizable, Data[ObjectDefinition], ABC):
```
bartfeenstra, line 78: Rename this to `_DateExpression`. Add `type Date expression = Date | Date range | _DateExpression`. The type will allow exhaustive type checking. The private base class allows code sharing. Including the base class in the type means the base class can type on `DateExpression` and be valid for itself and all subclasses.

betty/date.py
```diff
@@ -1,459 +1,491 @@
...
-            (date.year is not None, date.month is not None, date.day is not None)
-        ].localize(localizer)
-    except KeyError:
-        raise IncompleteDateError(
-            "This date does not have enough parts to be rendered."
-        ) from None
-    parts = (1 if x is None else x for x in date.parts)
-    return dates.format_date(datetime.date(*parts), date_parts_format, localizer.locale)
+    @final
+    def __init_subclass__(cls, **kwargs: Any):
```
bartfeenstra, line 84: Remove this after making the class private

PR #4514, opened 2026-07-10
Add `Maybe` and `Result` to encapsulate (missing) values and/or errors

### LINKED ISSUE #4513: Add a `Result` type
```python
class Result[ValueT, ErrorT: Exception](ABC):
  pass

@final
class Value(Result):
  pass

@final
class Error(Result):
  pass
```

Can we replace `Void` with this? Or if we need to keep it, refactor it onto `Result`?

Rename the existing `betty.functools.Result` to `ResultCapturer`, which returns a `Result` when called?

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/4513

## To do
- Document the new types
  - `Maybe`
    - Why not just `Nothing` (like why not just `Void`?). Why three types instead of one? Give a concrete example.
    - `Nothing` is to `None` what `NothingType` is to `NoneType`
    - Similar to null-safe operators in other languages.
  - `Result`
    - performance (include benchmark stats)
    - type safety
    - useful for validation: reuse exceptions for the details they convey, but not to raise them if and until necessary 
    - Similar functionality to null-safe operators in other languages
- Add `Maybe.optional -> SomethingT | None` and `betty.maybe.from_optional(maybe: None | SomethingT) -> Maybe[SomethingT]`
- Test that `Maybe` and `Result` work with match statements. We'll need to implement `__match_args__`. For results to be matchable on their error **type**, we may need to add `Error.cls` and expose it as a match arg.
- Test that `Result` works with exception groups
- Benchmark error mapping
- Fix `@todo`s

## Follow-ups
- Rename assertions to validators
- Add `type betty.validator.Invalid = Intersection[HumanFacingException, ValueError]`
- Add `type betty.validator.Validator[InputT, OutputT, ErrorT: Invalid] = Callable[[InputT], Result[OutputT, ExceptionGroup[ErrorT]]]`

## Benchmark
```python
from time import perf_counter_ns
from typing import Never

from betty.result import Error, Ok, Result


def ____raise_error() -> Never:
    raise ValueError("Hello, world!")


def ___raise_error() -> None:
    ____raise_error()


def __raise_error() -> None:
    ___raise_error()


def _raise_error() -> None:
    __raise_error()


def test_raise() -> None:
    try:
        _raise_error()
    except ValueError as error:
        # Ensure we use the exception in case the interpreter somehow lazily binds it.
        error.args


def _map_error(error: BaseException) -> Result:
    return Ok(error)


def test_result() -> None:
    Error(ValueError("Hello, world!")) ^ _map_error


iterations = 999999
tests = [
    test_raise,
    test_result,
]


def test() -> None:
    for test in tests:
        start = perf_counter_ns()
        for _i in range(iterations):
            test()
        end = perf_counter_ns()
        end - start
        if iterations:
            pass


if __name__ == "__main__":
    test()

```
With this setup, `test_raise()` is ~50% slower than `test_result()`. Raising exceptions becomes slower as the the stack trace grows, as per expectations.

### DIFF
--- a/betty/functools.py
+++ b/betty/functools.py
@@ -4,7 +4,6 @@ Provide functional programming utilities.
 
 from __future__ import annotations
 
-import contextlib
 import threading
 from asyncio import sleep
 from itertools import chain
@@ -20,8 +19,8 @@ from typing import (
 )
 
 from betty.asyncio import resolve_await
+from betty.maybe import Maybe, Nothing, Something
 from betty.threading import threadsafe
-from betty.typing import Void
 
 if TYPE_CHECKING:
     from collections.abc import Awaitable, Callable, Iterable, Iterator
@@ -119,72 +118,6 @@ def passthrough[T](value: T, /) -> T:
     Return the value.
     """
     return value
-
-
-def suppress[**P, T](
-    target: Callable[P, T], *exceptions: type[BaseException]
-) -> Callable[P, T | type[Void]]:
-    """
-    Return the value, but suppress any errors.
-    """
-
-    def _suppress(*target_args: P.args, **target_kwargs: P.kwargs) -> T | type[Void]:
-        with contextlib.suppress(*exceptions):
-            return target(*target_args, **target_kwargs)
-        return Void
-
-    return _suppress
-
-
-class ResultUnavailable(RuntimeError):
-    """
-    A :py:attr:`betty.functools.Result.result` is unavailable.
-    """
-
-    def __init__(self):
-        super().__init__(
-            "The result is unavailable because the target has not been called yet."
-        )
-
-
-@final
-class Result[**P, T]:
-    """
-    Decorate a callable and store its return value or raised exception.
-    """
-
-    __slots__ = "_error", "_result", "_target"
-    _error: BaseException
-    _result: T
-
-    def __init__(self, target: Callable[P, T], /):
-        self._target = target
-
-    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> T:
-        """
-        Call the target.
-        """
-        try:
-            self._result = self._target(*args, **kwargs)
-        except BaseException as error:
-            self._error = error
-            raise
-        else:
-            return self._result
-
-    def result(self) -> T:
-        """
-        Get the target's return value.
-
-        If the target raised an exception, calling this method will re-raise the exception.
-        """
-        try:
-            raise self._error
-        except AttributeError:
-            try:
-                return self._result
-            except AttributeError:
-                raise ResultUnavailable from None
 
 
 type DecoratorCallableType[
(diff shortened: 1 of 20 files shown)


betty/functools.py
```diff
@@ -135,7 +137,90 @@ def _suppress(*target_args: P.args, **target_kwargs: P.kwargs) -> T | type[Void]
...
+
+
+# @todo Besides the cases where Void can be replaced with Result and errors, in a few other places we use Void to indicate
+# @todo an initial state. How to support that? Oh, just Result[ | None, SomeErrorType] | None?
+# @todo
+# @todo A third use case is where we use Void in args/kwargs as an alternative ABSENT value to None where None would be a valid PRESENT value.
+# @todo
+# @todo
+@final
+class Nothing[ValueT, ErrorT: BaseException](_Result[ValueT, ErrorT]):
```
bartfeenstra: Add `Maybe` and make this `Nothing(Maybe)`

betty/functools.py
```diff
@@ -135,7 +137,90 @@ def _suppress(*target_args: P.args, **target_kwargs: P.kwargs) -> T | type[Void]
...
+    def value(self) -> ValueT:
+        raise self._error
+
+    @override
+    @property
+    def error(self) -> ErrorT:
+        return self._error
+
+
+type Result[ValueT, ErrorT: BaseException] = (
```
bartfeenstra: Document this. Explain this is for deferred error handling, and that catching exceptions is slow. Explain that one of the use cases is validation (where failures are expected), allowing detailed error informatie to be returned in the form of an unraised exception. 

Errors encoded in `Result` are also typed (as opposed to raised exceptions), which should improve DX.

betty/typing.py
```diff
@@ -52,17 +57,224 @@ def threadsafe[T](target: T, /) -> T:
...
+    @abstractmethod
+    def on_value[NewValueT](
+        self, function: Callable[[ValueT], NewValueT], /
+    ) -> Result[NewValueT, ErrorT]:
+        """
+        Apply a function to the result value, and return its return value.
+        """
+
+    @abstractmethod
+    def on_error[NewErrorT: BaseException](
```
bartfeenstra: Should we allow an error type to be specified so the function is only called for errors of that type?

betty/typing.py
```diff
@@ -52,17 +57,224 @@ def threadsafe[T](target: T, /) -> T:
     return target
 
 
+class _Result[ValueT, ErrorT: BaseException](ABC):
+    @property
+    @abstractmethod
+    def ok(self) -> ValueT:
```
bartfeenstra: May also return `Never`

betty/result.py
```diff
@@ -0,0 +1,193 @@
...
+# @todo This is really not specific to the Result API.
+# @todo Refactor this into a generic solution to decorate a callable and store its result for later access.
+# @todo
+# @todo Except.... We may need this for easy function application/pipelines. Maybe test that first before
+# @todo we remove/refactor this class?
+# @todo
+# @todo
+# @todo
+@final
+class FromCall[**P, ReturnT, ErrorT: BaseException]:
```
bartfeenstra: Make this more generally usable and rename it `Snapshot`?

PR #4588, opened 2026-07-21
Automatically localize `Localizable`s before printing them in Jinja templates

### LINKED ISSUE #4587: Templates should localize content automatically
`jinja2.compiler.CodeGenerator.visit_Output()` is responsible for parsing `{{ }}` statements, so to speak. We can use our own code generator that overrides this method (or `CodeGenerator._output_child_pre()` and `CodeGenerator._output_child_post()`) to add a call around the evaluated value (but before any escaping) to a method on our generator, that then checks if the value is of a certain type and must be processed before returning, or if it must be returned directly.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/4587

## To do
- Remove `html_lang` entirely? The problem with this, is that it seems it may not be possible to reliably and automatically determine if and when a `LocalizedStr` needs to be printed in an HTML element, or when not.
- Remove as many `localize` calls from templates as possible.
- To do that, we may need to automate the application of the `html_lang` filter as well
- And to do **that**, we need to solve it not only for when printing on its own, but also when a localizable is printed as 1) the value of an HTML attribute (we cannot use our `<span>` trick then) and 2) the contents of an HTML tag that does not allow `<span>` inside of it.
- Consider `{% element %}` in addition to or instead of the newly introduced `element` global

### DIFF
--- a/betty/jinja/__init__.py
+++ b/betty/jinja/__init__.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 
 import datetime
 from asyncio import to_thread
-from collections.abc import Awaitable, Callable, Iterable
+from collections.abc import Awaitable, Callable, Iterable, Mapping
 from os import makedirs
 from pathlib import Path
 from shutil import copy2
@@ -29,6 +29,7 @@ from betty.locale import default_locale, to_language_tag
 from betty.localizable import Localizable, ResolvableLocalizable
 from betty.localizables.gettext import gettext, ngettext, npgettext, pgettext
 from betty.localizables.markup import JoinAnd, JoinOr
+from betty.localized import LocalizedStr
 from betty.machine_name import MachineName
 from betty.media_type import (
     ResolvableMediaType,
@@ -36,6 +37,7 @@ from betty.media_type import (
     match_extension,
     resolve_media_type,
 )
+from betty.media_types.html import HTML
 from betty.media_types.jinja import JINJA
 from betty.pathlib import resolve_path
 from betty.store import StoreItem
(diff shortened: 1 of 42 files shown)


betty/jinja/__init__.py
```diff
@@ -62,6 +65,20 @@ def context_document(context: JinjaContext) -> Document:
     return document
 
 
+@final
+class _GlobalGettext[*Ts]:
```
bartfeenstra, line 71: This not specific to gettext. Rename to `_LocalizableFactory`

betty/jinja/__init__.py
```diff
@@ -253,3 +240,51 @@ async def _cache(
...
+
+@pass_context
+def _new_html_attributes(
+    context: Context, **attributes: Unpack[AttributesKwargs]
+) -> Attributes:
+    return Attributes(localizer=context_document(context).localizer).set(**attributes)
+
+
+@pass_context
+def _html_tag(
```
bartfeenstra, line 322: s/tag/element

bartfeenstra, line 322: Move this under `betty.html`

bartfeenstra, line 322: Support void elements? See https://developer.mozilla.org/en-US/docs/Glossary/Void_element

betty/jinja/__init__.py
```diff
@@ -117,6 +135,7 @@ async def new_environment(project: Project, /) -> Environment:
         "secondary_navigation_links": [
             link.link for link in project.links if not link.primary
         ],
+        "tag": _html_tag,
```
bartfeenstra, line 140: s/tag/element 

Should this instead be an extension and use an `{% element %}` tag? It would make longer content easier to manage.

betty/tests/asset_directories/raspberry_mint/templates/component/test_accordion.py
```diff
@@ -49,7 +49,7 @@ async def test_with_html_attributes(assert_template_file: AssertTemplateFile) ->
                     "body": Markup("<p>Lorem ipsum dolor sit amet</p>"),
                 },
             ],
-            "attributes": Attributes(html_class=[html_class]),
+            "attributes": Attributes().set(html_class=[html_class]),
```
bartfeenstra, line 52: Undo this, as the Localizer argument does not clash with any of the html kwargs

PR #4590, opened 2026-07-21
Refactor translations

bartfeenstra:


betty/gettext.py
```diff
@@ -174,59 +167,110 @@ def _find_source_files(
                 yield input_file
 
 
-class TranslationRepository(ABC):
+class Translations(ABC):
```
bartfeenstra, line 167: Do we even need this base class anymore? All the translation methods are gettext-specific anyway, and they are duplicated on `Localizer`, which is the entry point for other code to call.

betty/content_builders/raspberry_mint_incomplete_translation_warning.py
```diff
@@ -27,12 +39,46 @@ class IncompleteTranslationWarning(Template, Manufacturable):
...
     async def build_template(self, document: Document) -> TemplateBuild:
-        return "component/raspberry-mint/incomplete-translation-warning.html.j2"
+        completion = 0
+        if document.localizer.locale in self._completions:
+            translated, total = self._completions[document.localizer.locale]
+            if translated == total:
+                return None
+            completion = round(100 * translated / total)
+        return "component/raspberry-mint/incomplete-translation-warning.html.j2", {
+            "translation_completion": completion,
```
bartfeenstra: Prefix this variable with the **component** name.

betty/gettext.py
```diff
@@ -173,188 +156,148 @@ def _find_source_files(
...
-            else:
-                for entry in pofile(po):
-                    if entry.translated():
-                        yield entry.msgid_with_context
+    return _builtin_translation_locales
 
 
-@final
-class UntranslatedLocale(LocaleError):
+async def load_translations(
```
bartfeenstra: Refactor this to load the translations for a specific locale only.

betty/project/__init__.py
```diff
@@ -515,28 +515,29 @@ async def generate_entity_list_html(
         )
 
     @service
-    async def translations(self) -> TranslationRepository:
+    def translations(self) -> TranslationsRepository:
```
bartfeenstra: Do we really need the translations repositories to be public services? If we just inline this in the localizer repository services, we remove a complexity. Then for isolated apps and projects we can just inject localizer repositories without translations repositories and we won't have to worry about needless I/O anymore.

betty/localizer.py
```diff
@@ -26,13 +26,16 @@ class Localizer:
...
     def __init__(
-        self, locale: ResolvableLocale, translations: gettext_api.NullTranslations, /
+        self, locale: ResolvableLocale, /, *, translator: Translator | None = None
     ):
         self.locale: Final[Locale] = resolve_locale(locale)
         """
         The locale.
         """
-        self._translations = translations
+        self.translator: Final[Translator] = translator or Translator()
```
bartfeenstra: Rename to `.translate`

PR #4607, opened 2026-07-23
Improve console output

### LINKED ISSUE #3264: Add helpers to `User` to check if certain behavior is enabled
Add helpers to `User` to check if certain behavior is enabled, e.g. to evaluate the verbosity to check if warnings are emitted at all. 

Add default implementations so this does not break BC. Use the new functionality in subclasses. 

Update existing code that compares `User.verbosity` explicitly to use this instead.

### LINKED ISSUE #4187: Add `User.with_verbosity()`
## To do
- It should return a new instance with its own verbosity
- Use it to replace the current way to change the verbosity

### LINKED ISSUE #4605: Add a console argument to show all exception tracebacks
Add a console argument to show all exception tracebacks, because it looks really bad when this happens when not explicitly requested.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/4605, https://github.com/bartfeenstra/betty/issues/4187, and https://github.com/bartfeenstra/betty/issues/3264

## To do
- Add test coverage for the new traceback argument


betty/user/__init__.py
bartfeenstra: Require an exception argument. Add another final method that does not take arguments, but calls this method with the exception currently being handled.

PR #4626, opened 2026-07-29
Add a search API

bartfeenstra:
This aims to refactor the existing search code into a public API, independent of themes, built around data definitions and without hardcoded special cases. This means that we won't change existing search features (much) at this time, but make existing features available to any and all entity types, and other types of content.

## To do
- Can we inject projects into factories?
- Can we use locks to allow lazy capability initialization, and to remove stages entirely? Refactor this in a separate PR.
- We see errors like `betty.capability.NotYetInitialized: betty.datas.optional:OptionalDefinition's "indexer" capability was not yet initialized for stage betty.definition.cls:OnSetCls` appear. It turns out it's a bit tricky to proxy indexers (because we have two different types) lazily in `OptionalDefinition`.
  We should probably allow capability manufacturers to return `None`?
- How to index iterables?
- Indexing needs to return a locale along with the text. Search implementations can then enable the correct tokenization, such as for CJK. See also https://www.fusejs.io/token-search.html
  For most accurate tokenization, we must be able to extract the parts of a `LocalizedStr` in each of the locales (e.g. if the localized string contains a placeholder value in a different locale). Fuse.js lets us use custom tokenizers on the source data where we can process it per locale into `String[]`
  This is an enhancement that's useful elsewhere as well (e.g. when printing `LocalizedStr` in HTML templates, and adding `lang` attributes), so do this in a separate PR.
- Consider using a library for the front-end:
  - runs in-browser (WASM/JS/TS)
  - can precompile the index
  - can parse queries with field selectors and filters and conditions, as well as API-based query construction at the same time (so we can add the search form filters to the parsed query)
  - field weight/boost/importance
  - some kind of middleware so we can expand and finetune the search behavior in the future, e.g. with language-specific improvements (stop words, stemming, etc)
  - Fuse.js?
- To improve initial page performance, only load the library JS and the index if and when the search UI is activated/interacted with. 
  - Add a concurrency API with an asynchronous lock. 
  - Add a `LocalizedStr` equivalent in TS
- Fix `@todo`s


betty/search.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class Searcher(ABC):
+    """
+    Search a single resource.
+    """
+
+    @classmethod
+    @abstractmethod
+    def fields(cls) -> Mapping[str, FieldDefinition]:
```
bartfeenstra: Because we have `ResolvableSearcher`, we really don't want to declare fields here. Given a searchable definition, we want to get the field definitions.

bartfeenstra: However, it also makes sense to keep field definitions and the indexing code together. Can't have one without the other.

data/webpack/search/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@betty.py/betty.extensions.raspberry-mint",
```
bartfeenstra: Update this file

data/webpack/search/main.ts
```diff
@@ -0,0 +1,5 @@
+"use strict"
+
+import { Search } from "./search.ts"
```
bartfeenstra, line 3: Merge both files into one

betty/search.py
```diff
@@ -0,0 +1,137 @@
...
+    @abstractmethod
+    async def index(
+        self, data: DataT, /, *, localizer: Localizer, context: Context | None
+    ) -> Index:
+        """
+        Index the data.
+        """
+
+    @abstractmethod
+    async def render_result(
```
bartfeenstra, line 99: This is also only needed for 'top-level' data

betty/entity/__init__.py
```diff
@@ -106,6 +109,102 @@ async def linked_data_schema(cls, project: Project, /) -> JsonLdObject:
...
+            "entity.id": entity.id,
+            "entity.label": await self._index_label(entity, localizer=localizer),
+            "entity.description": await self._index_has_description(
+                entity, localizer=localizer
+            ),
+            "entity.notes": await self._index_has_notes(entity, localizer=localizer),
+        }
+
+    async def _index_label(self, entity: EntityT, /, *, localizer: Localizer) -> str:
+        return entity.label.localize(localizer)
```
bartfeenstra: Only do this if there is a label **attr**, because this method returns the generated label, derived from other data (an attr, or formatted from the entity ID, for example)

betty/data.py
```diff
@@ -56,6 +59,10 @@ def __init__(
             Intersection[PorterT, Porter[DataT]], Self, DataT
         ]
         | None = None,
+        indexer: ResolvableDataDefinitionFeature[
```
bartfeenstra: In a separate PR:
- rename this to "(cap)abilities" 
- add (cap)abilities to `FieldDefinition`

bartfeenstra: Make this a separate API with a few tools that code can use? 
- Refactor `DataDefininitionFeature` and move it into this API
- Add tools to support arbitrary abilities. 
- Add 'stages' for when ability factories need to be run. This is to support factories that depend on `ClsDefinition`, where the class may be set after the definition is initialized.

betty/data.py
```diff
@@ -113,6 +122,26 @@ def try_porter(self) -> Intersection[PorterT, Porter[DataT]] | None:
...
+    @final
+    @property
+    def indexer(self) -> IndexerT:
+        """
+        The search indexer for the data.
+
+        :raises betty.search.NotSearchable:
+        """
+        if not self._indexer:
+            raise NotSearchable("This data is no searchable.")
```
bartfeenstra: "not"

PR #4630, opened 2026-08-10
Allow `OwnerAttr` and `CollectionOwnerAttr` to be frozen

### LINKED ISSUE #4548: Read-only attributes
One of the main reasons not all of our code uses attributes is that there is currently no way to make them read-only once a value has been set. However, some objects are immutable, and this prevents us from refactoring them onto our new components.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/4548

### DIFF
--- a/betty/classtools.py
+++ b/betty/classtools.py
@@ -4,7 +4,13 @@ Tools to create classes.
 
 from __future__ import annotations
 
-from typing import Self, final, override
+from abc import ABCMeta
+from functools import cache
+from inspect import getmembers
+from typing import TYPE_CHECKING, Any, Self, final, override
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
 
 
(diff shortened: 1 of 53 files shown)


betty/classtools.py
```diff
@@ -20,3 +21,45 @@ def __new__(cls):
...
+    def __call__(cls, *args: Any, **kwargs: Any):
+        new = cls.__new__(cls, *args, **kwargs)
+        new._pre_init()
+        new.__init__(*args, **kwargs)
+        new._post_init()
+        new.__is_initialized = True
+        return new
+
+
+class Init(metaclass=_InitMeta):
```
bartfeenstra: Rename to `Initializable`, or ideally something even more descriptive.

Expand the docstring, and explain the class' purpose is "cooperative initialization"

Issue #4203, opened 2026-05-15
Add `Freezable`


bartfeenstra:
Now that all of our data will be based around data definitions, and in practice most will be data objects with properties, we can support mutability for entities and configuration by adding support to attributes. 

## August 2026 thoughts
- No mutability, but freezing (based on experience with https://github.com/bartfeenstra/betty/pull/4630)
- Add `Freezy`/`MaybeFrozen`/`?`as a base class
  - abstract method property `.is_frozen`
- Add `Freezable(Freezy)` which allows a one-time operation to make an unfrozen object frozen. This is for data types that must remain editable until fully set up, after which they are made immutable.
  - `__slots__`
  - `.freeze()`, which is idempotent, and a no-op after the first call.
- Make `Frozen` subclass `Freezy`
- `assert_frozen()`
- `assert_not_frozen()`
- Update `is_frozen()` to check for `Freezy` as well, or only?
- Make `Data` subclass `Freezy`
- `freeze()` which takes one or two arguments (`Data`, or `Any` and `DataDefinition`). There is another API function like this somewhere. Ensure we use consistent argument order. Do we want two versions of this: one that tries to freeze as much as possible (for production) and one that raises an error if it finds data that is not frozen or freezable (for testing, so we can ensure all built-in data/configuration is or can be made frozen)?
- Make our concrete collections subclass `Freezy`
- Freeze the ancestry in the load API
- Freeze data objects after loading them from portable data:
  - app configuration 
  - project configuration 

## To do'
- Use https://github.com/bartfeenstra/betty/pull/4252 for inspiration
- Use 0.4.x for inspiration: https://github.com/bartfeenstra/betty/blob/0.4.x/betty/mutability.py
- Add `betty.mutable.Mutable`
  - Add `betty.mutable.MutabilityError(Exception)` and `betty.attr.ImmutableAttr(MutabilityError, Owner error)`
  - `Mutable.mutable` is a `@property` whose setter only allows making the owner mutable, never immutable.
  - Add `Mutable.assert_mutable()` that raises `MutabilityError`
  - Add `Attr.assert_mutable(owner)` that raises `ImmutableAttr`
- Add `HasAttrs(HasProperties, Mutable, Data[ObjectDefinition])`
- Expand `Attr` with owner mutability checks. 
- Add explicit `AttributeError` subclasses
- Make ancestries immutable in `betty.load.load()`
- Make configuration immutable once loaded, e.g. `AppData` and `ProjectData`

PR #4635, opened 2026-08-11
Make the `isolated_project` test fixture session-scoped

### LINKED ISSUE #4567: Improve performance of tests that use projects
E.g. template tests (see https://github.com/bartfeenstra/betty/issues/4400). `TestDeriver` is a good benchmarking subject, because it does a lot of small things, and it requires a project but does not really use it.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/4567

## Blocked by 
- https://github.com/bartfeenstra/betty/issues/4203

### DIFF
--- a/betty/test_utils/conftest.py
+++ b/betty/test_utils/conftest.py
@@ -8,6 +8,7 @@ to start using these utilities.
 from __future__ import annotations
 
 __all__ = [
+    "_isolated_project",
     "assert_dumps_linked_data",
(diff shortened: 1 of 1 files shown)


bartfeenstra:
Or replace it with a `frozen_project` fixture once we can freeze data. 

Or simply have one `app` and one `project` fixture that each return isolated, frozen service levels?

PR #4653, opened 2026-08-17
Decouple services from service levels

bartfeenstra:


betty/services/plugin/__init__.py
```diff
@@ -304,18 +311,20 @@ async def _collect_requirement(
...
+):
     """
     A plugin service provider.
     """
 
     def __init__(
         self,
         *args: Any,
-        services: ServiceLevel,
+        services: ServiceLevelT,
```
bartfeenstra, line 328: Rename this so it is namespaced: `plugin_service_services`

bartfeenstra: Or introduce `HasPrivateServiceLevel` which adds the kwarg but sets `self._services`

Issue #4128, opened 2026-04-27
Allow asynchronous `@service` factories to be called during service provider bootstrap, making services themselves synchronous


bartfeenstra:
## To do
- Make this an optional part of `ServiceManager` under the hood, so it can be used for any individual service.
- for services marked as such, allow asynchronous factories as well
- Do we want to extract this into a new base class that integrates `Prop` and the life cycle API? How will that relate to `PluginServiceManager.init_plugins()`?
- That means these services will not need a lock
- implement this for `AssetRepository`
- Implement this for `Project.localizer`

bartfeenstra:
What would we use this for? Reopen if/when we have concrete use cases.

bartfeenstra:
Ah, yes, for services that depend on plugin services, for example. Make it `@service(auto=True)`?

PR #4666, opened 2026-08-24
Improve service factories

### LINKED ISSUE #4673: Remove `ServiceManager.override()`
Can we replace it with calls to `Prop.set()` instead? Now that we have `Object` and `Type`, we can keep track of initialization status anyway.

Also, somewhat related, once https://github.com/bartfeenstra/betty/pull/4666 has been merged, can service resolvers not replace the owner attr with the resolved value? They use locks anyway, so could safely mutate the data.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/4673

## To do
- See what we can do about https://github.com/bartfeenstra/betty/issues/4128 for simple service managers
- Ideally we allow that for any service manager: regardless of whether the manager returns a sync or an async service, it may be overridden with any sync or async manufacturer.
- Fix `@todo`s

## Blocked by
- https://github.com/bartfeenstra/betty/pull/4683


betty/service.py
```diff
@@ -16,12 +16,37 @@
...
+# @todo
+# @todo Also, if we are to use these types, we must use ServiceLevelT: ServiceLevel instead of OwnerT: HasServices
+# @todo for the argument we pass on to the manufacturers?
+# @todo But they would also benefit from access to the owner. Which may not itself have services or a service level (data definitions).
+# @todo What if manufacturers can take one (service level) or two (service level and owner) arguments? (or no arguments at all)
+# @todo
+# @todo
+type _ServiceManufacturerReturn[ServiceT] = ServiceT | Coroutine[Any, Any, ServiceT]
+type ServiceManufacturer[
+    OwnerT: HasServices,
```
bartfeenstra: If we inject service levels separately, owners no longer have to subclass any specific type, opening up the way for mixin types.

bartfeenstra: It would also support data definitions as owners.

betty/service.py
```diff
@@ -16,12 +16,37 @@
...
+    | Callable[[ServiceLevelT, OwnerT], _ServiceManufacturerReturn[ServiceT]]
 )
+type ServiceOrManufacturer[
+    OwnerT: HasServices,
+    ServiceT,
+    ServiceLevelT: ServiceLevel = ServiceLevel,
+] = Service[ServiceT] | ServiceManufacturer[OwnerT, ServiceT, ServiceLevelT]
 
 
 class HasServices[ServiceLevelT: ServiceLevel = ServiceLevel](
```
bartfeenstra: In a separate PR, rename and move this

betty/service.py
```diff
@@ -16,12 +16,37 @@
...
+# @todo What if manufacturers can take one (service level) or two (service level and owner) arguments? (or no arguments at all)
+# @todo
+# @todo
+type _ServiceManufacturerReturn[ServiceT] = ServiceT | Coroutine[Any, Any, ServiceT]
+type ServiceManufacturer[
+    OwnerT: HasServices,
+    ServiceT,
+    ServiceLevelT: ServiceLevel = ServiceLevel,
+] = (
+    Callable[[], _ServiceManufacturerReturn[ServiceT]]
```
bartfeenstra: Can we instead of repeating Callable, use a type alias that takes a type var tuple?

betty/service.py
bartfeenstra: Maybe also add `ResolvableServiceLevel` from that other PR

bartfeenstra: Ensure that is implemented for requirements too

betty/service.py
```diff
@@ -5,25 +5,45 @@
...
 from betty.typing import Intersection
 
 if TYPE_CHECKING:
-    from betty.service_level import ResolvableServiceLevel
+    from betty.service_level import ResolvableServiceLevel, ServiceLevel
 
-type ServiceFactory[OwnerT: ResolvableServiceLevelHasServices, FactoryServiceT] = (
-    Callable[[OwnerT], FactoryServiceT]
+
+type _ServiceManufacturerReturn[ServiceT] = ServiceT | Coroutine[Any, Any, ServiceT]
```
bartfeenstra: A `Optional*` variants of all type aliases

bartfeenstra: Not needed, since `ServiceT` can include `None` as well.

betty/service.py
```diff
@@ -5,25 +5,49 @@
...
-    Callable[[OwnerT], FactoryServiceT]
+type _ServiceManufacturerReturn[ServiceT] = ServiceT | Coroutine[Any, Any, ServiceT]
+
+
+type ServiceManufacturer[
+    OwnerT: ResolvableServiceLevelHasServices,
+    ServiceT,
+    ServiceLevelT: ServiceLevel = ServiceLevel,
+] = (
+    Callable[[], _ServiceManufacturerReturn[ServiceT]]
```
bartfeenstra: Separate each callable signature into a separate type, so we can use those with `cast()` in `call_manufacturer()` and get rid of the ty:ignore comments

betty/service.py
```diff
@@ -5,25 +5,49 @@
...
+type ServiceOrManufacturer[
     OwnerT: ResolvableServiceLevelHasServices,
     ServiceT,
-    FactoryServiceT,
-] = Service[ServiceT] | ServiceFactory[OwnerT, FactoryServiceT]
+    ServiceLevelT: ServiceLevel = ServiceLevel,
+] = Service[ServiceT] | ServiceManufacturer[OwnerT, ServiceT, ServiceLevelT]
+
+
+type TypedServiceOrManufacturer[
```
bartfeenstra: Move support for typed services to a submodule? Also add scaffolding to make calling code more readable.

betty/service.py
```diff
@@ -5,25 +5,49 @@
...
 if TYPE_CHECKING:
-    from betty.service_level import ResolvableServiceLevel
+    from betty.service_level import (
+        ResolvableServiceLevel,
+        ServiceLevel,
+    )
 
-type ServiceFactory[OwnerT: ResolvableServiceLevelHasServices, FactoryServiceT] = (
-    Callable[[OwnerT], FactoryServiceT]
+type _ServiceManufacturerReturn[ServiceT] = ServiceT | Coroutine[Any, Any, ServiceT]
```
bartfeenstra: `_ManufacturerReturn`

betty/service.py
```diff
@@ -5,25 +5,49 @@
...
+    ServiceLevelT: ServiceLevel = ServiceLevel,
+] = (
+    Callable[[], _ServiceManufacturerReturn[ServiceT]]
+    | Callable[[ServiceLevelT], _ServiceManufacturerReturn[ServiceT]]
+    | Callable[[ServiceLevelT, OwnerT], _ServiceManufacturerReturn[ServiceT]]
 )
-type ServiceOrFactory[
+
+
+type ServiceOrManufacturer[
```
bartfeenstra: Replace `ServiceOrManufacturer` with `ServiceInit` and `service_or_manufacturer` with `service_init` everywhere.

betty/service.py
bartfeenstra: Or keep here, but add `resolve_service_init()`?

betty/objecttools.py
```diff
@@ -27,11 +27,23 @@ def has(self, object_: ObjectT, /) -> bool:
         """
         return hasattr(object_, self.name)
 
+    @overload
     def get(self, object_: ObjectT, /) -> Any:
+        pass
+
+    @overload
+    def get(self, object_: ObjectT, default: Any, /) -> Any:
```
bartfeenstra, line 35: Add test coverage

betty/services/simple.py
```diff
@@ -0,0 +1,55 @@
...
+    from betty.asyncio import ReAwaitable
+    from betty.service import ResolvableServiceLevelHasServices, ServiceManager
+
+
+@overload
+def service[OwnerT: ResolvableServiceLevelHasServices, ServiceT](
+    manufacturer: Callable[[OwnerT], Coroutine[Any, Any, ServiceT]],
+    /,
+    *,
+    sync: bool = False,
```
bartfeenstra, line 35: Add test coverage

betty/service.py
```diff
@@ -5,110 +5,198 @@
...
-    ServiceT,
-    FactoryServiceT,
-] = Service[ServiceT] | ServiceFactory[OwnerT, FactoryServiceT]
+] = (
+    Manufacturer[ServiceT, ServiceLevelT]
+    | _OwnerDependentServiceManufacturer[ServiceT, ServiceLevelT, OwnerT]
+)
+
+
+async def new[OwnerT: ResolvableServiceLevelHasServices, ServiceT: ServiceType](
```
bartfeenstra, line 56: Order the type vars so the return value comes first

betty/service.py
```diff
@@ -5,110 +5,198 @@
...
+
+
+type WrappableServiceInit[
+    ServiceT: ServiceType,
+    ServiceLevelT: ServiceLevel,
+    OwnerT: ResolvableServiceLevelHasServices,
+] = ServiceT | ServiceInit[OwnerT, ServiceT, ServiceLevelT]
+
+
+type OptionalWrappableServiceInit[
```
bartfeenstra, line 117: Perhaps this is where a `Maybe` type would be useful, as it would allow services to be of **any** type.

bartfeenstra: The issue with that, however, is that it would no allow us to use classes as manufacturers , since they return the final value rather than a container. They could be wrapped , but that would hurt DX and readability.

betty/service.py
```diff
@@ -5,110 +5,198 @@
...
-    GetServiceT,
-    GetterServiceT,
-    FactoryServiceT,
-](Prop[OwnerT, GetServiceT]):
+    ServiceT: ServiceType,
+    GetT,
+    ResolverT,
+](Prop[OwnerT, GetT, ServiceInit[ServiceT, ServiceLevel, OwnerT] | NothingType]):
     """
     Manage a single service for a service provider.
```
bartfeenstra, line 132: Update the docstring. Service providers have come to mean something different since this was written.

betty/service.py
bartfeenstra: We should no longer need `betty.service.new()` once https://github.com/bartfeenstra/betty/pull/4683 has been merged

betty/service.py
```diff
@@ -5,110 +5,191 @@
...
-type ServiceOrFactory[
+
+
+type _ManufacturerReturn[ServiceT: ServiceType] = (
+    ServiceT | Coroutine[Any, Any, ServiceT]
+)
+
+
+type _OwnerDependentServiceManufacturer[
+    ServiceT: ServiceType,
```
bartfeenstra, line 40: Once https://github.com/bartfeenstra/betty/pull/4683 has been merged, refactor this onto that, but also only use one arg, and that being `ResolvableServiceLevel`.

PR #4667, opened 2026-08-24
Add `ResolvableServiceLevel`

bartfeenstra:


betty/service_level.py
```diff
@@ -97,3 +101,32 @@ def __init__(
         """
         The upstream service level.
         """
+
+
+@disjoint_base
```
bartfeenstra: This class is meant to be a mixin for **any** other class. As such it should not have any bases, and none of the classes in here should be made disjoint? 

For `ResolvableServiceLevel`, can we instead define it as `Intersection[ServiceLevel, Not[HasServiceLevel]] | Intersection[HasServiceLevel, Not[ServiceLevel]]`?

PR #4683, opened 2026-08-27
Add a callback API to allow the factory API and arbitrary callbacks to pass on variable arguments

bartfeenstra:
## Thursday Thoughts
- Report on the different kinds of classes that currently subclass `*Manufacturable`
- All this binding is great for callbacks, but a lot of faff for classes.
- simpler, and faster, would be to use base classes for specific purposes, like we do now.
- Different sets of allowed base classes for different APIs
- **NOT** for data plugins. They don't have manufacturers (or if they do, they aggregates and their definitions are factories.
- For plugins, add `betty.plugin.factory.new()` which also accepts regular callables **that take ALL provided args**
- Or.... We standardize, and all factories and callbacks always take a single argument, which is a carrier. The base type can then be subclassed.

## To do
- "manufacturer" is a long word. "factory" is not. Can we rename manufacturers back to factories?
- Now ty has been upgraded to a version that supports type var tuples, refactor this API to leverage those
  - Remove the hardcoded arg limit `max_arg_count`
  - Merge `Arg*Manufacturable` into `Manufacturable`
  - Simplify the tests? 
- High level API:
  - call with any number of `AnyCallable`
  - call with any number of `CompiledCallable`
  - bind values to any number of `AnyCallable`
  - bind values to any number of `CompiledCallable`
  - And in the future once we can check types against types:
    - bind types to any number of `AnyCallable`
    - bind types to any number of `CompiledCallable`
- Tests
  - Test our support for overloaded callbacks.
  - Type safety tests
- What to do with `DataManufacturable`?
- Fix `@todo`s

## Follow-ups
- Can we use this for multiple dispatch/overloading?


betty/factory.py
```diff
@@ -4,69 +4,123 @@
...
+    @classmethod
+    @abstractmethod
+    async def new(cls) -> Self:
+        """
+        Create a new instance.
+        """
 
 
-class Manufacturable(metaclass=ABCMeta):
+class Arg1Manufacturable[ServiceLevelT: ServiceLevel](metaclass=ABCMeta):
```
bartfeenstra: Can we make the whole API service level agnostic? `factory.new(manufacturer, factory.services)` would do the same thing.

bartfeenstra: At that point, do we really need arg3 support at this time?

betty/factory.py
```diff
@@ -4,121 +4,359 @@
...
+        self.manufacturer: Final[Any] = manufacturer
 
-class ManufacturerError(FactoryError):
+
+class InvalidManufacturer(ManufacturerError, TypeError):
+    """
+    Raised if a value would never be a valid manufacturer under any circumstances.
+    """
+
+    def __init__(self, manufacturer: Any, reason: str, /):
```
bartfeenstra: Rename to `value`, because it is in fact not a manufacturer

betty/factory.py
```diff
@@ -4,121 +4,359 @@
...
+        )
+
+
+class UnsupportedManufacturer(ManufacturerError, ValueError):
     """
-    Raised when a manufacturer raised an error while creating a new object.
+    Raised when a manufacturer is not supported for the given :py:func:`new() <betty.factory.new>` args.
     """
 
+    def __init__(self, manufacturer: Any, new_args: tuple[Any, ...], reason: str, /):
```
bartfeenstra: Type on `Arg2Manufacturer`

betty/tests/test_factory.py
```diff
@@ -1,275 +1,422 @@
...
-            _async_callable_manufacturer_without_services_with_required_arguments,
-            _async_callable_manufacturer_with_services_with_required_arguments,
-        ],
-    )
-    async def test_new__should_raise_unsupported_manufacturer(
-        self, manufacturer: Manufacturer
-    ) -> None:
-        with pytest.raises(UnsupportedManufacturer):
-            await Factory(ServiceLevel()).new(manufacturer)
+@pytest.mark.parametrize(
```
bartfeenstra: Expand this with classes as well

betty/factory.py
bartfeenstra: Also in subclasses

betty/factory.py
```diff
@@ -4,121 +4,346 @@
...
+from typing import Any, Final, Self, final, overload
+
+from typeguard import TypeCheckError, check_type
 
 from betty.asyncio import resolve_await
-from betty.data import Data
-from betty.importlib import fully_qualified_name
-from betty.service_level import ServiceLevel
-from betty.typing import Intersection
+from betty.localizables.markup import JoinOr
```
bartfeenstra: Remove the dependency on localizables. Maybe add `betty.string.join_or()` (and `betty.string.join_and()`)?

betty/factory.py
```diff
@@ -4,121 +4,374 @@
...
+    """
+    Raised when a manufacturer is not callable.
+    """
+
+    def __init__(self, manufacturer: Any, /):
+        super().__init__(manufacturer, "it is not callable")
+
+
+@final
+class UnevaluatedManufacturerArgType(InvalidManufacturer):
```
bartfeenstra: This error now indicates an invalid manufacturer. However, manufacturers may have untyped arguments, as long as we don't need them. That means this must subclass `UnsupportedManufacturer`

bartfeenstra: And moved below with the other subclasses

betty/data/factory.py
```diff
@@ -0,0 +1,26 @@
...
+from __future__ import annotations
+
+from abc import abstractmethod
+
+from betty.data import Data
+from betty.factory import Arg2Manufacturable
+from betty.service_level import ServiceLevel
+
+
+class DataManufacturable[ServiceLevelT: ServiceLevel, DataT: Data](
```
bartfeenstra, line 14: Can we decouple this from service levels?

betty/factory.py
```diff
@@ -4,121 +4,92 @@
...
+        """
+        Create a new instance.
+        """
 
 
-class ManufacturerError(FactoryError):
+class Arg1Manufacturable[Arg1T](metaclass=ABCMeta):
     """
-    Raised when a manufacturer raised an error while creating a new object.
+    Allow this type to be initialized using a argument.
```
bartfeenstra: "an"

betty/factory.py
```diff
@@ -4,121 +4,92 @@
...
 
-type Manufacturer[T] = (
-    type[Intersection[T, Manufacturable]]
-    | Callable[[], Coroutine[Any, Any, T] | T]
-    | Callable[[ServiceLevel], Coroutine[Any, Any, T] | T]
+type Manufacturer[T] = type[Manufacturable] | Callback[T]
+
+
+type Arg1Manufacturer[T, Arg1T] = (
+    type[Arg1Manufacturable[Arg1T]] | Arg1Callback[T, Arg1T]
```
bartfeenstra, line 56: And `Manufacturable`

betty/call.py
```diff
@@ -0,0 +1,334 @@
...
+            return callback, _match_callback(callback, *args)
+        except* CallbackError as error:
+            errors.extend(error.exceptions)
+    raise ExceptionGroup(
+        f"Could not match {join_or(*map(repr, callbacks))} to the given new args.",
+        errors,
+    )
+
+
+def _validate_callback(callback: AnyCallback, /) -> tuple[Parameter, ...]:
```
bartfeenstra: Make this public so we can focus the tests on a small portion of behavior.

betty/call.py
```diff
@@ -0,0 +1,334 @@
...
+        return await return_value
+    return return_value
+
+
+def _expand_callbacks(callback: AnyCallback, /) -> Iterable[AnyCallback]:
+    # @todo Handle overloads and forward references
+    yield callback
+
+
+def _match_callbacks(
```
bartfeenstra: Make this public so we can focus the tests on a small portion of behavior.

betty/call.py
```diff
@@ -0,0 +1,334 @@
...
+                parameter.kind is Parameter.KEYWORD_ONLY
+                and parameter.default is Parameter.empty
+            ):
+                raise RequiredCallbackKwarg(callback, parameter)
+    except InvalidCallback as error:
+        raise ExceptionGroup("Invalid callback", [error]) from None
+    return parameters
+
+
+def _match_callback(callback: AnyCallback, *args: Any) -> tuple[Any, ...]:
```
bartfeenstra: Make this public so we can focus the tests on a small portion of behavior.

betty/call.py
```diff
@@ -0,0 +1,334 @@
...
+            _assert_type(callback, args, parameter, args[parameter_number])
+            matched_args.append(args[parameter_number])
+        elif parameter.kind is Parameter.VAR_POSITIONAL:
+            _assert_type(callback, args, parameter, args[parameter_number:])
+            matched_args.extend(args[parameter_number:])
+            break
+    return tuple(matched_args[:arg_count])
+
+
+def _assert_type(
```
bartfeenstra: Make this public so we can focus the tests on a small portion of behavior.  Also rename it to `assert_value()`.

betty/call.py
```diff
@@ -0,0 +1,334 @@
+"""
+The call API.
```
bartfeenstra: Rename to callback API?

betty/call.py
```diff
@@ -0,0 +1,334 @@
...
+        super().__init__(
+            callback,
+            args,
+            f'it has an arg `{parameter.name}` whose type hint "{parameter.annotation}" is unevaluated',
+            parameter,
+        )
+
+
+@final
+class IncompatibleCallbackArg(UnsupportedCallbackArg):
```
bartfeenstra: Rename to `IncompatibleCallbackArgType`.

betty/call.py
```diff
@@ -0,0 +1,334 @@
...
+        super().__init__(
+            callback,
+            args,
+            f"it has a required arg `{parameter.name}`, and not enough new args to be able to map one to it",
+            parameter,
+        )
+
+
+@final
+class UntypedCallbackArg(UnsupportedCallbackArg):
```
bartfeenstra: Rename to `MissingCallbackArgType`.

betty/call.py
```diff
@@ -0,0 +1,334 @@
...
+        parameter: Parameter,
+        /,
+    ):
+        super().__init__(callback, args, reason, parameter=parameter)
+
+
+@final
+class RequiredCallbackArg(UnsupportedCallbackArg):
+    """
+    Raised when a callback has a required arg, and there are not enough new args to be able to map one to it.
```
bartfeenstra: s/map/match everywhere

betty/call.py
```diff
@@ -0,0 +1,334 @@
...
+    """
+    Call a callback and return its return value.
+
+    :param args: Any arguments to pass on to the callback, if it accepts them.
+
+    :raises CallError:
+    :raises InvalidCallback:
+    :raises SupportedCallback:
+    """
+    # @todo 1) Expand all callbacks (e.g. replace by overloads)
```
bartfeenstra: Finish this

PR #4713, opened 2026-09-11
Move most of the factory API into the classed plugin API

### LINKED ISSUE #4675: Move `DataManufacturable` out of the factory API

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/4675

## Goals
- Remove argument matching based on argument names and types.
- Remove the hardcoded coupling with service levels and make it optional 
- Remove `DataManufacturable` and move the configuration class declaration to plugin definitions, to fully decouple concrete plugin classes from the plugin API.
- Improve DX by bringing back the term "configuration" for defined data in the context of plugin manufacturing.

## To do
- Update the commit message and PR title to reflect the changes in scope: refactor the factory API into smaller, decoupled APIs.
- Allow manufacturable plugins to set a custom factory on their definitions. This supports third-party classes as plugins, but also pavea the way for decoupling plugin classes from other system components, and centralizing integrationa in definitions.
- Refactor `betty.service_level.factory.new()` into `ServiceLevel.new()`
- Refactor `betty.plugin.cls.factory.new()` into `ClassedPluginDefinition.new()`
- Add `IntegratablePluginDefinition` with its own `.new()`
- Refactor `betty.plugin.config.factory.new()` into `ConfigurablePluginDefinition.new()`
- We want to start supporting plugins on service providers, so the `services` argument should be typed `ResolvableServiceLevel`.
- We currently feed data plugins through the factory API, but we shouldn't, really. They have no configuration (they are defined data themselves), and should never integrate with any services.
- we can group our plugins as follows:
  1) metadata only (unclassed plugins, definitions only)
  2) data plugins (classed, portable date data support, no async factories)
  3) functional plugins (classed and/or configurable, async factories and integration)
- If config is not relevant (for groups 1 and 2, or those plugins from group 3 that take no or optional config), the minimum we would need as a factory is `ResolvablePluginDefinition`. `NewPlugin` is really only ever relevant for those factories that want to use config. And even then we could do without it (but would we want to?) if the plugin ID and config are all fully typed (and not portable data).
- all 3 groups MUST be creatable from portable data:
  1) a plugin ID only
  2) a plugin ID and (almost always) the plugin's dumped data (although not all plugins may have fields other than their ID).
  3) a plugin ID and for some plugins their dumped configuration.
  
  Groups 2 and 3 require similar portable data minimization, although the source of the data and factory process are different.
- Group 2 must be recreatable from the same source data (portable or plugin instances?), yielding fully unique instances after every creation. Group 3 MAY receive this support (there is no reason to stop it from inheriting it, for instance), but no use case for that has been identified so far.
- Having said all that, the `WikipediaContributors` copyright notice is a wonderful exception to the above: it's a data plugin, but its data depends on an HTTP response (needed to determine which Wikipedia copyright URL to link to), `ProjectAuthor` is another, with a much simpler integration dependency: it just needs the project author, something that does not even need to be async (but there is no hurt in it being async).
  Can we move this async dependency injection into plugin discoveries? They are async already, and have access to the relevant service level. Except then these plugins would not be referenceable by code as their classes would be generated (even though we know there will always be exactly one of each). We could abstract that away by adding `NewWikipediaContributors` and `NewProjectAuthor` that hardcode the string plugin IDs, but so calling code is type-safe?
- Add `betty.plugin.data.factory`?
- Support non-`type` callables as factories?


betty/plugin/cls.py
```diff
@@ -4,17 +4,50 @@
...
+    ResolvablePluginDefinition,
+    ResolvablePluginId,
+    ResolvablePluginTypeDefinition,
+    resolve_plugin_id,
+    resolve_plugin_type_definition,
+)
+from betty.portable import KeyedPorter, PortableData
+from betty.prop import HasProps
+from betty.sample import Samplable, Sample, Samples, Size
+from betty.service_level import Integratable, ServiceLevel
```
bartfeenstra, line 44: Move the factory code (for all APIs) back into `factory` submodules. Factory APIs are integration-heavy and moving the relevant code into separate modules will save on imports.

betty/plugin/cls.py
```diff
@@ -47,4 +82,341 @@ def _set_cls(self, cls: type[PluginT], /) -> None:
...
+        super().__init__(*args, **kwargs)
+        self.__configuration_cls = configuration_cls
+
+    @property
+    def configuration_cls(self) -> type[Data]:
+        """
+        The plugin's configuration class, if it is configurable.
+        """
+        if not self.__configuration_cls:
+            raise NotConfigurable
```
bartfeenstra, line 188: Provide args

betty/plugin/cls.py
```diff
@@ -47,4 +82,341 @@ def _set_cls(self, cls: type[PluginT], /) -> None:
...
+
+    @final
+    async def __call__(self, services: ServiceLevel, /) -> PluginT:
+        """
+        Create a new instance of the configured plugin.
+        """
+        plugin_cls = (await services.plugins[self.data().plugin_type][self.id]).cls
+        args = (plugin_cls, services)
+        if self.configuration is not Nothing:
+            if not issubclass(plugin_cls, ConfigurablePlugin.__value__):
```
bartfeenstra, line 329: Use the plugin definition

betty/plugin/cls.py
bartfeenstra: Similarly, add a `new()` function to every API's `factory` submodule. `betty.plugin.cls.factory.new()` can then call the others.

PR #4729, opened 2026-09-19
Add a definition API

### LINKED ISSUE #4724: Add `HasDefinition`
Make it a base class that replaces `Plugin.plugin()` and `Data.data()` with a `.definition` read-only class attribute. Can we use a descriptor to accomplish this? 

Add `Definition`.

bartfeenstra:
This fixes https://github.com/bartfeenstra/betty/issues/4724

## To do
- Extract plugin IDs into `IdentifiableDefinition`?
- Fix `@todo`s


betty/definition/cls.py
```diff
@@ -67,3 +78,24 @@ def cls(self) -> type[BaseClsT] | None:
...
+    ) -> DefinitionT:
+        return self._definition
+
+
+class HasDefinition[DefinitionT: _ClsDefinition](metaclass=ABCMeta):
+    """
+    A class that exposes its definition.
+    """
+
+    definition: ClassVar[_HasDefinitionGetter[DefinitionT]]
```
bartfeenstra: The getter does not have to be generic itself. It can get the relevant typevar from the owner in its `.__get__()` like we do with `Prop`.

